PR #12, opened 2022-09-27
Give coin by reacting to messages

MahdiBM:
This will enable giving others coin, by only reacting to their messages.
Currently it'll only work with 🪙 and vapor server's own coin emoji.


MahdiBM:
I did think about adding the heart as well. 
The only thing is that the heart emojis are used a lot even if someone doesn't want to give a coin, so even if liking a post justifies a coin, this might (or might not) prove a bit more confusing than it should.
The coin emojis though, seem more explicit. People usually don't use the emojis as a reaction, so if they do its likely to explicitly give a coin.

So what do we do? 
Do we want to keep it more explicit or make it more lax? We can always change the coin signs in the future if it proves problematic.

MahdiBM:
I had initially considered writing tests, but i another took a look to try to see what we can write tests for.

The only thing I can see that we can write unit tests for is `ReactionCache`, but it gets complicated because there is a `DiscordClient` call in `ReactionCache` which we can't easily mock. Generally speaking `ReactionCache` doesn't do anything complicated so i don't know if it's worths writing tests for.

How should i approach this?

MahdiBM:
Another thing .. I guess it might be better to add the heart emojis too ... we can always remove them if they prove problematic.

PR #16, opened 2022-10-07
Better responses

MahdiBM:
### Problem
Currently the user-responses are made in the user-service, which limits our ability to customize them based on the context that user-service is unaware of.
For example it's not possible to customize the response if the coin was given using a reaction.

### Modifications
I've done some changes to what is returned from user-service.
Now it just returns some info, and doesn't make the response upfront.

I also did some general code-cleanup.

### DIFF
--- a/CODE/Sources/PennyAPI/AddCoin/AddCoinLambdaHandler.swift
+++ b/CODE/Sources/PennyAPI/AddCoin/AddCoinLambdaHandler.swift
@@ -64,18 +64,23 @@ struct AddCoins: LambdaHandler {
                 id: UUID(),
                 createdAt: Date(),
                 amount: product.amount,
-                from: userUUID,
+                from: .init(
+                    userID: userUUID,
+                    discordID: product.from
+                ),
                 source: product.source,
                 reason: product.reason)
             
-            let message = try await userService.addCoins(with: coinEntry, to: user)
-            response = APIGatewayV2Response(statusCode: .ok, body: message)
+            let coinResponse = try await userService.addCoins(with: coinEntry, to: user)
+            let data = try JSONEncoder().encode(coinResponse)
+            let string = String(data: data, encoding: .utf8)
+            response = APIGatewayV2Response(statusCode: .ok, body: string)
         }
         catch UserService.ServiceError.failedToUpdate {
             response = APIGatewayV2Response(statusCode: .notFound, body: "ERROR- The user in particular wasn't found.")
         }
         catch let error {
-            response = APIGatewayV2Response(statusCode: .badRequest, body: "ERROR-\(error.localizedDescription)")
+            response = APIGatewayV2Response(statusCode: .badRequest, body: "ERROR- \(error.localizedDescription)")
             
         }
         return response
--- a/CODE/Sources/PennyAPI/AddCoin/AddCoinLambdaHandler.swift
+++ b/CODE/Sources/PennyAPI/AddCoin/AddCoinLambdaHandler.swift
@@ -64,14 +64,15 @@ struct AddCoins: LambdaHandler {
                 id: UUID(),
                 createdAt: Date(),
                 amount: product.amount,
-                from: .init(
-                    userID: userUUID,
-                    discordID: product.from
-                ),
+                from: userUUID,
                 source: product.source,
                 reason: product.reason)
             
-            let coinResponse = try await userService.addCoins(with: coinEntry, to: user)
+            let coinResponse = try await userService.addCoins(
+                with: coinEntry,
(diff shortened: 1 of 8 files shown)


MahdiBM:
I'll ask benny for a review too, to make sure I haven't changed a part of the app the shouldn't have been changed.
Initially I changed `CoinEntry` and was passing some extra info in it, but then I realized it's going to be saved on the db too (?), and I need to be more careful about changing it.

CODE/Sources/PennyAPI/AddCoin/AddCoinLambdaHandler.swift
```diff
@@ -68,14 +68,20 @@ struct AddCoins: LambdaHandler {
...
+            let data = try JSONEncoder().encode(coinResponse)
+            let string = String(data: data, encoding: .utf8)
+            response = APIGatewayV2Response(statusCode: .ok, body: string)
         }
         catch UserService.ServiceError.failedToUpdate {
             response = APIGatewayV2Response(statusCode: .notFound, body: "ERROR- The user in particular wasn't found.")
         }
         catch let error {
-            response = APIGatewayV2Response(statusCode: .badRequest, body: "ERROR-\(error.localizedDescription)")
+            response = APIGatewayV2Response(statusCode: .badRequest, body: "ERROR- \(error.localizedDescription)")
```
BennyDeBock, line 84: Is there supposed to be a space here?

CODE/Sources/PennyBOT/Handlers/MessageHandler.swift
```diff
@@ -46,33 +46,30 @@ struct MessageHandler {
             )
             do {
                 let response = try await self.coinService.postCoin(with: coinRequest)
-                if response.starts(with: "ERROR-") {
```
BennyDeBock, line 49: A thought I've had, but maybe we could send the errors to the bot-lab channel so we don't always have to dive into the logs unless the bot crashes. I have no idea if that's possible though

BennyDeBock:
I've left some comments, for the rest it looks good to me. `CoinEntry` does indeed get saved into the DB as a user has multiple coin entries.

CODE/Sources/PennyAPI/AddCoin/AddCoinLambdaHandler.swift
MahdiBM: I just looked at the log that was above it and it had space so I thought maybe this should have a space too 😅. Otherwise I guess it doesn't really matter?

CODE/Sources/PennyBOT/Handlers/MessageHandler.swift
MahdiBM: Personally I've kind of liked looking at the logs in AWS, it's convenient to me and you can expand JSONs and look at them properly.

For logging these to a channel in Discord, I personally use a Discord logger in my other project which uses Web hook to send data to a Discord channel.
I've been meaning to make it publicly available with proper configurations, but im not sure when i'll get to it.

We can add something like that if we really want to, just using normal create-message or a Web hook.

CODE/Sources/PennyAPI/AddCoin/AddCoinLambdaHandler.swift
BennyDeBock: Ah, mistake on my end then haha. But it doesn't matter that much tbf, we should just be consistent, that's all

CODE/Sources/PennyBOT/Services/CoinService.swift
```diff
@@ -2,32 +2,35 @@ import Foundation
...
-            
-        } catch {
-            return "request failed: \(error)"
+    func postCoin(with coinRequest: CoinRequest) async throws -> CoinResponse {
+        let BASE_URL = ProcessInfo.processInfo.environment["API_BASE_URL"]
+        logger.trace("\(String(describing: BASE_URL))")
+        var request = HTTPClientRequest(url: "\(BASE_URL ?? "")/coin")
+        request.method = .POST
+        request.headers.add(name: "Content-Type", value: "application/json")
+        request.body = .bytes(Data(try! JSONEncoder().encode(coinRequest)))
```
0xTim, line 21: Let's remove the force unwrap here, _just in case_

0xTim (approved):
One nit otherwise looks good

CODE/Sources/PennyBOT/Services/CoinService.swift
MahdiBM: yeah I saw that afterwards too. 
I don't know how that could fail, but at the same time the function is throwing already, so no need to risk anything.

PR #21, opened 2022-10-14
Add More Tests

MahdiBM:
### Problem
Penny needs better tests so we can put more trust in it to be working after each update.

### Some Explanation
I don't intend to add a lot of tests in this PR. 
What I want to do is to prepare the abstractions and clear up the way.
Then anybody will be able to write some tests without a need to worry about more complex stuff.


MahdiBM:
Swift nightly tests seem to due to one of the usual sendable changes.

@0xTim feel free to let me know if any improvements can be done.

CODE/Sources/PennySHARED/Repositories/DynamoUserRepository.swift
```diff
@@ -3,47 +3,37 @@ import SotoDynamoDB
...
-    case itemNotFound
-}
-
-public struct DynamoUserRepository: UserRepository {
+struct DynamoUserRepository: UserRepository {
     
     // MARK: - Properties
     let db: DynamoDB
     let tableName: String
-    let eventLoop: EventLoop
```
0xTim, line 19: We should bring this back - the event loop from the DynamoDB instance isn't necessarily the same event loop as the context it's being called from. It's probably fine for now but making it explicit ensures we don't shoot ourselves in the foot in the future

CODE/Tests/Fake/FakeAWSHTTPClient.swift
```diff
@@ -0,0 +1,36 @@
+import SotoCore
+import NIOCore
+import NIOHTTP1
+import Logging
+import Foundation
+
+public class FakeAWSHTTPClient: AWSHTTPClient {
```
0xTim, line 7: Isn't this protocol disappearing?

CODE/Tests/Fake/FakeManager.swift
```diff
@@ -0,0 +1,176 @@
...
+            }
+        } else {
+            return
+        }
+    }
+    
+    public func send(key: EventKey) {
+        let data = testData(for: key.rawValue)!
+        let decoder = JSONDecoder()
+        let event = try! decoder.decode(Gateway.Event.self, from: data)
```
0xTim, line 78: It would be nice not to crash the tests on error here

0xTim (approved):
A few comments but no blockers, great work!

CODE/Tests/Fake/FakeAWSHTTPClient.swift
MahdiBM: o hmm, I don't follow AWS SDKs closely, but I can see that they're removing it: https://github.com/soto-project/soto-core/commit/addcfe6fd27b05e5fedb83ebbd29d90223100432

I think we might need to make a "AWSClientProtocol" of ourselves now?

MahdiBM: oo the thing is we're still using soto 5 instead of the newest soto 6 ... that's why they are removing the protocol (since it no longer has any use in soto 6)

0xTim: We can update Soto in a separate PR. Where are you abstracting the AWSClient? (I'm on mobile). We probably want to move the abstraction a layer up rather to eg a EC2Interface rather than the client directly. That will make it easier to migrate Soto versions too

MahdiBM: So that'll be a "AWSClientProtocol" of ourselves again? This way we can hide the AWS interface under a protocol of ourselves and only add functions we need (which makes updating easier too)

0xTim: Yes though I think mocking at the client level is too low and it should be abstracted higher. (Think the UserRepository level instead of mocking DynamoDB directly)

MahdiBM: about "where?" its in the "add coin lambda handler" (PennyAPI/AddCoin)

MahdiBM: The AWSClient is also used in `PennySHARED/Services` (btw we should restructure the folders and targets too 😅 in another pr)

So should i create a new target "AWSClientWrapper"?

MahdiBM: Sorry it appears Github doesn't update the conversation unless you refresh the page, so i missed your last comment.

MahdiBM: Replying to your last comment:
Yeah i was thinking of that too a while ago (although i forgot about it later)...  i'll see what i can do

MahdiBM:
Its interesting that the test randomly throw 
> note: module 'PennyBOT' is the main module of an executable, and cannot be imported by tests and other targets

once in a while, even in Swift 5.6 ...
I'll check to see if something can be done ... I've seen some other weird test failures with swift 5.6 too ...

0xTim:
Why not just drop 5.6 and move to 5.7?

MahdiBM:
we definitely should 😅
should _i_ do it? i can update the main branch and merge this pr with main so we can test the pr with 5.7

0xTim:
Yep sounds good

CODE/Tests/Fake/FakeAWSHTTPClient.swift
MahdiBM: I didn't find a better solution than the `underlyingHandler` solution.
We can simplify these _after_ we update our soto stuff, including the runtime.

MahdiBM: Ok i just noticed something ... let me check ...

MahdiBM: Pushed some more changes to simplify those stuff

MahdiBM:
All ready!

CODE/Tests/Fake/FakeManager.swift
```diff
@@ -0,0 +1,184 @@
...
+    func respond(to endpoint: Endpoint, with payload: Any) {
+        if let continuation = continuations.removeValue(forKey: endpoint.urlSuffix) {
+            continuation.resume(returning: payload)
+        } else {
+            unhandledResponses[endpoint.urlSuffix] = payload
+        }
+    }
+}
+
+private struct FakeDiscordClient: DiscordClient {
```
0xTim, line 147: It would be nice to split these out into their own types

0xTim (commented):
One last comment then good to go

CODE/Tests/Fake/FakeManager.swift
MahdiBM: I assume you mean the `send` and `sendAndAwaitResponse ` functions?

0xTim: Sorry I meant files - `FakeDiscordClient` and `FakeResponseStorage` should probably live in their own files

MahdiBM: good idea :\)

0xTim (approved):

PR #26, opened 2022-11-04
Introducing `BotStateManager` to avoid duplicate answers by Penny on startup

MahdiBM:
### Problem
When we update Penny, AWS waits a few minutes before taking down the old Penny instance to make sure the new instance is healthy.
This makes it so there is a short period where there are 2 Penny bots that will respond to Discord Gateway events.
Only the new bot should respond to events.

### Solution
This PR introduces a new `BotStateManager` actor.
On initialization, this actor sends a message to an internal channel in Discord that Penny has access to, and asks the other Pennys to stop responding to new gateway events.
The old Pennys upon receiving this message, will stop responding to gateway events for the next 3 minutes.
If the new instance is healthy, AWS will shutdown the old instance after 1-2-3 minutes. Otherwise the old instance will start processing Gateway events again after those 3 minutes, and the bot won't be disabled.

Other than having some tests, I've also manually tested this PR.

### DIFF
--- a/CODE/Package.swift
+++ b/CODE/Package.swift
@@ -6,7 +6,7 @@ import PackageDescription
 let package = Package(
     name: "PennyAPI",
     platforms: [
-        .macOS(.v12)
+        .macOS(.v13)
     ],
     products: [
         .executable(name: "PennyLambdaAddCoins", targets: ["PennyLambdaAddCoins"]),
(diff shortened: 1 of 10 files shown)


0xTim (approved):
👍 let's see if this works

PR #31, opened 2022-11-10
Added better error descriptions

ptoffy:
This improves error descriptions to make debugging easier

### DIFF
--- a/CODE/Sources/PennyAPI/Sponsors/SponsorLambda.swift
+++ b/CODE/Sources/PennyAPI/Sponsors/SponsorLambda.swift
@@ -197,7 +197,7 @@ struct AddSponsorHandler: LambdaHandler {
         
         // Throw if adding new role response is invalid
         guard 200...299 ~= removeRoleResponse.status.code else {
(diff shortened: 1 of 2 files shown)


CODE/Sources/PennyAPI/Sponsors/SponsorLambda.swift
```diff
@@ -183,7 +197,7 @@ struct AddSponsorHandler: LambdaHandler {
         
         // Throw if adding new role response is invalid
         guard 200...299 ~= removeRoleResponse.status.code else {
-            context.logger.error("Failed to remove \(role.rawValue) role from user \(userDiscordID) with error: \(removeRoleResponse.status.description)")
+            context.logger.error("Failed to remove \(role.rawValue) role from user \(userDiscordID) with error: \(removeRoleResponse.status.description) and body: \(removeRoleResponse.body?.debugDescription ?? "empty")")
```
0xTim, line 200: This is a `ByteBuffer` right? We should convert it to a string properly as I don't think description will actually print the body

0xTim (changes requested):

0xTim (approved):

PR #42, opened 2023-03-08
Auto-Pings lambda

MahdiBM:


CODE/task-def.json
```diff
@@ -28,10 +28,6 @@
                 {
                     "name": "BOT_APP_ID",
                     "value": "950695294906007573"
-                },
-                {
-                    "name": "PINGS_API_BASE_URL",
-                    "value": "CHANGE_ME.com"
```
0xTim, line 34: Do we not need this?

0xTim (approved):
This looks good to me. I did check and we're building the lambda and uploading in CI so should be good. Happy to merge and fix forward if needed

CODE/task-def.json
MahdiBM: You know better, but i came to the conclusion that the pings lambda will use the same api-base-url? Like all lambdas use that url, but depending on the path parameters, aws would redirect the requests to the appropriate lambda function based on what we defined in the Stack.yml file?

PR #43, opened 2023-03-18
Structural changes + Better README + More documentation

MahdiBM:
Also updated the CI files. Might have missed something but should be good enough to test-deploy.

Current tree:
```
└─ .
   ├─ .DS_Store
   ├─ CODE
   │  ├─ .DS_Store
   │  ├─ Dockerfile
   │  ├─ Makefile
   │  ├─ Package.resolved
   │  ├─ Package.swift
   │  ├─ Sources
   │  │  ├─ .DS_Store
   │  │  ├─ PennyAPI
   │  │  │  ├─ AddCoin
   │  │  │  │  └─ AddCoinHandler.swift
   │  │  │  ├─ AutoPings
   │  │  │  │  └─ AutoPingsHandler.swift
   │  │  │  ├─ Helps
   │  │  │  │  └─ HelpsHandler.swift
   │  │  │  └─ Sponsors
   │  │  │     ├─ GithubWebhookPayload.swift
   │  │  │     ├─ SponsorLambda.swift
   │  │  │     └─ SponsorType.swift
   │  │  ├─ PennyBOT
   │  │  │  ├─ +Array.swift
   │  │  │  ├─ +DiscordColor.swift
   │  │  │  ├─ +Logger.swift
   │  │  │  ├─ .DS_Store
   │  │  │  ├─ BotStateManager.swift
   │  │  │  ├─ CommandsManager.swift
   │  │  │  ├─ Constants.swift
   │  │  │  ├─ DiscordFactory.swift
   │  │  │  ├─ Handlers
   │  │  │  │  ├─ +Expression.swift
   │  │  │  │  ├─ +String.swift
   │  │  │  │  ├─ CoinFinder.swift
   │  │  │  │  ├─ EventHandler.swift
   │  │  │  │  ├─ InteractionHandler.swift
   │  │  │  │  ├─ MessageHandler.swift
   │  │  │  │  └─ ReactionHandler.swift
   │  │  │  ├─ Penny.swift
   │  │  │  ├─ ProposalsChecker.swift
   │  │  │  └─ Services
   │  │  │     ├─ AutoPingsService
   │  │  │     │  ├─ AutoPingsService.swift
   │  │  │     │  └─ DefaultPingsService.swift
   │  │  │     ├─ CoinService
   │  │  │     │  ├─ CoinService.swift
   │  │  │     │  └─ DefaultCoinService.swift
   │  │  │     ├─ DiscordService
   │  │  │     │  └─ DiscordService.swift
[...]

PR #44, opened 2023-04-15
/help command

### LINKED ISSUE #20: Commands responding with pre-defined responses
Penny should be able to respond with certain responses when it detects a pre-defined command.
This is very similar to #19, but this should only work when used at the beginning of a message or line so it doesn't spam users with unrelated message just because they used a frequently-used word.

MahdiBM:
resolves #20

### Checklist
* [x] Implement the functionality in Penny
* [x] Add a "Helps" lambda
* [x] Add the lambda/bucket to the configuration files
* [x] Some tests
* [ ]  [after the pr is approved] Add manual lambda/bucket stuff to aws console

It doesn't matter if this is merged before #43 ... I kind of prefer this to be merged first 🤔


CODE/Sources/PennyAPI/Helps/HelpsHandler.swift
```diff
@@ -0,0 +1,82 @@
+import AWSLambdaRuntime
+import AWSLambdaEvents
+import Foundation
+import SotoCore
+import PennyModels
+import PennyRepositories
+import PennyExtensions
+import NIOConcurrencyHelpers
+
```
MahdiBM, line 9: The lambda. Works pretty similar to AutoPings

CODE/Sources/PennyBOT/CommandsManager.swift
```diff
@@ -10,7 +10,7 @@ struct CommandsManager {
 
     private func makeCommands() -> [Payloads.ApplicationCommandCreate] {
```
MahdiBM, line 11: In main-branch commits i've done some clean ups around how we handle Application/Slash commands.
Penny now uses enums to cleanly register commands to Discord, and also cleanly decode them in `InteractionHandler` when receiving from Discord.

CODE/Sources/PennyBOT/CommandsManager.swift
```diff
@@ -53,7 +56,16 @@ enum SlashCommand: String, CaseIterable {
...
-                return .init(
+                ApplicationCommand.Option(
+                    type: .subCommand,
+                    name: subCommand.rawValue,
+                    description: subCommand.description,
+                    options: subCommand.options
+                )
+            }
+        case .help:
+            return HelpSubCommand.allCases.map { subCommand in
```
MahdiBM, line 67: The rest of the file is just help command declarations

CODE/Sources/PennyBOT/Constants.swift
```diff
@@ -35,5 +35,12 @@ enum Constants {
             .moderator,
             .core,
         ]
+
+        static let elevatedRestrictedCommandsAccess: [Roles] = [
```
MahdiBM, line 39: Help command add/remove is only accessible to these roles.
We can also put a setting using Discord clients so these are enforced by Discord too.

CODE/Sources/PennyBOT/Handlers/+String.swift
```diff
@@ -5,7 +5,7 @@ import PennyModels
 /// `StringProtocol` is basically either `String` or `Substring`.
 extension StringProtocol {
     /// trims whitespaces and makes the string case, diacritic and punctuation insensitive.
-    func foldForPingCommand() -> String {
+    func heavyFolded() -> String {
         self.trimmingCharacters(in: .whitespaces)
```
MahdiBM, line 9: renamed this since it is now used in /help too

CODE/Sources/PennyBOT/Handlers/InteractionHandler.swift
```diff
@@ -16,6 +16,9 @@ struct InteractionHandler {
     var pingsService: any AutoPingsService {
         ServiceFactory.makePingsService()
     }
+    var helpsService: any HelpsService {
+        ServiceFactory.makeHelpsService()
```
MahdiBM, line 20: This whole file is adding command-handling logic for /help.
You would need a good understanding about how Discord's complicated slash commands work, so i don't really expect you to actually review this file. You can still try to review the each command's logic, but that would probably be troublesome too. I've tested all these using a test app. If something is not working as expected i'll just fix it.

CODE/Sources/PennyBOT/Handlers/InteractionHandler.swift
```diff
@@ -522,7 +711,12 @@ private enum ModalID {
         case add, remove, test
```
MahdiBM, line 711: This `ModalID` type is also to decode/encode modal-ids (`custom_id`) easier. Makes stuff much cleaner.
The Modals penny uses are just those "Text Editor"s like in /auto-pings

CODE/Sources/PennyBOT/Handlers/InteractionHandler.swift
```diff
@@ -706,3 +881,72 @@ extension Payloads.InteractionResponse.Modal: Response {
     /// Responses containing a modal can't be an edit to another message.
```
MahdiBM, line 881: I've also recently introduced this `Response` type. The command-responding function can return any type that conforms to `Response`, then the top-level send-message function would automatically know what to send where.
Didn't turn out to be the cleanest  approach, but works for now.

CODE/Sources/PennySHARED/Repositories/S3HelpsRepository.swift
```diff
@@ -0,0 +1,77 @@
+import SotoS3
+import Foundation
+import PennyModels
+import PennyExtensions
+
+struct S3HelpsRepository: HelpsRepository {
```
MahdiBM, line 6: Again, mostly a copy-paste of what we already have working in AutoPings

CODE/Tests/PennyBOTTests/GatewayProcessingTests.swift
```diff
@@ -365,6 +367,50 @@ class GatewayProcessingTests: XCTestCase {
         )
         XCTAssertTrue(message.hasSuffix(" \(Constants.vaporCoinEmoji)!"))
     }
+
+    func testHelpsCommand() async throws {
```
MahdiBM, line 371: 4 basic-functionality tests for the new /help

MahdiBM (commented):
Self-review

MahdiBM:
(If you're wondering, i'm trying to update this branch every time there is a change in main to avoid having to deal with a massive merge conflict)

CODE/Sources/PennyBOT/Handlers/InteractionHandler.swift
```diff
@@ -277,6 +296,46 @@ private extension InteractionHandler {
...
             }
+        case let .help(helpMode):
+            switch helpMode {
+            case .add:
+                let name = try modal.components
+                    .requireComponent(customId: "name")
+                    .requireTextInput()
+                    .value.requireValue()
+                    .trimmingCharacters(in: .whitespaces)
+                    .replacingOccurrences(of: "\n", with: " ")
```
gwynne, line 307: Do we actually want to permit help-text names with newlines in them, or is this a matter of Discord being too liberal with line breaks?

CODE/Sources/PennyBOT/Handlers/InteractionHandler.swift
```diff
@@ -388,6 +444,107 @@ private extension InteractionHandler {
...
+            let member = try event.member.requireValue()
+            guard await discordService.memberHasRolesForElevatedRestrictedCommandsAccess(
+                member: member
+            ) else {
+                let rolesString = Constants.Roles
+                    .elevatedRestrictedCommandsAccess
+                    .map(\.rawValue)
+                    .map(DiscordUtils.mention(id:))
+                    .joined(separator: " ")
+                return "You don't have access level for this command. This command is only available to \(rolesString)"
```
gwynne, line 487:
```suggestion
                return "You don't have access to this command; it is only available to \(rolesString)"
```

CODE/Sources/PennyBOT/Handlers/InteractionHandler.swift
```diff
@@ -388,6 +444,107 @@ private extension InteractionHandler {
...
+            ) else {
+                /// To not make things too complicated for now, send an acknowledgment,
+                /// so the String we return can be sent as an "edit".
+                guard await sendAcknowledgement(isEphemeral: true) else { return nil }
+                let rolesString = Constants.Roles
+                    .elevatedRestrictedCommandsAccess
+                    .map(\.rawValue)
+                    .map(DiscordUtils.mention(id:))
+                    .joined(separator: " ")
+                return "You don't have access level for this command. This command is only available to \(rolesString)"
```
gwynne, line 514:
```suggestion
                return "You don't have access to this command; it is only available to \(rolesString)"
```

CODE/Sources/PennyBOT/Handlers/InteractionHandler.swift
```diff
@@ -388,6 +444,107 @@ private extension InteractionHandler {
...
+            .getAll()
+            .filter({ $0.key.heavyFolded().contains(foldedName) })
+            .sorted(by: { $0.key < $1.key })
+        /// Discord only allows sending a max of 25 choices
+        if choices.count > 25 {
+            choices = Array(choices.dropLast(choices.count - 25))
+        }
+        return Payloads.InteractionResponse.Autocomplete(
+            choices: choices.map { .init(name: $0.key, value: .string($0.value)) }
+        )
```
gwynne, line 546:
```suggestion
        return try await Payloads.InteractionResponse.Autocomplete(choices: helpsService
            .getAll()
            .filter { $0.key.heavyFolded().contains(foldedName) }
            .sorted { $0.key < $1.key }
            .prefix(25)
            .map { .init(name: $0.key, value: .string($0.value)) }
        )
```

CODE/Sources/PennyBOT/Services/DiscordService/DiscordService.swift
```diff
@@ -376,6 +376,12 @@ actor DiscordService {
         })
     }
 
+    func memberHasRolesForElevatedRestrictedCommandsAccess(member: Guild.Member) -> Bool {
+        Constants.Roles.elevatedRestrictedCommandsAccess.contains(where: {
+            member.roles.contains($0.rawValue)
+        })
```
gwynne, line 382:
```suggestion
        !(Set(member.roles).intersection(
            Constants.Roles.elevatedRestrictedCommandsAccess.map(\.rawValue)
        )).isEmpty
```

CODE/Sources/PennySHARED/Repositories/S3HelpsRepository.swift
```diff
@@ -0,0 +1,77 @@
...
+    func remove(name: String) async throws -> [String: String] {
+        var all = try await self.getAll()
+        if all.removeValue(forKey: name) != nil {
+            try await self.save(items: all)
+        }
+        return all
+    }
+
+    func getAll() async throws -> [String: String] {
+
```
gwynne, line 36:
```suggestion
```

CODE/Tests/PennyBOTTests/GatewayProcessingTests.swift
```diff
@@ -388,6 +390,50 @@ class GatewayProcessingTests: XCTestCase {
...
+            }
+        }
+
+        do {
+            let response = try await manager.sendAndAwaitResponse(
+                key: .helpsAddFailure,
+                as: Payloads.EditWebhookMessage.self
+            )
+            let message = try XCTUnwrap(response.embeds?.first?.description)
+            XCTAssertTrue(message.hasPrefix("You don't have access level for this command. This command is only available to"), message)
```
gwynne, line 413: If you use my suggested grammar for this message, update this test

gwynne (changes requested):
A couple of actual concerns, otherwise mostly just style nits and cleanliness stuff.

CODE/Sources/PennyBOT/Handlers/InteractionHandler.swift
MahdiBM: Discord allows new lines, so here I'm just replacing those new lines with a simple space incase someone actually sent a multi-line name.

CODE/Sources/PennyBOT/Services/DiscordService/DiscordService.swift
MahdiBM: Is it worth going through the whole hash-map creation just to perform the containment check? 🤔

gwynne: Yes indeedy. It's actually faster this way even for tiny arrays - believe me, I've checked.

Proof:

1. `Set.init(_:)` and `Collection.map(_:)` are both `O(n)`, as is `Set.intersection(_:)` where the input is a `Sequence`. Big-O factors out coefficients, so the entire thing is effectively `O(n)`.
2. `Collection.contains(where:)` is `O(n)`, so nesting it yields `O(n*m)` (where `m` represents the second array). Big-O also disregards such differences (being somwhat similar to a derivative), which gives you `O(n^2)` complexity.

Even for linear versus polynomial complexity, Are you gonna notice the difference when the inputs are arrays of sizes 4 and 6 (just for example)? Very unlikely, since string comparison usually reduces to `memcmp()` when the length check doesn't short-circuit in modern Swift. But, the nested-contains thing is an _EXTREMELY_ common anti-pattern I see people using all the time, and quite often in cases where it makes a _VERY_ large difference (for example, say the inner loop was executing database queries on each iteration - which I've run into several times in production code!). It is absolutely worth the effort even if just to get yet another example of what not to do out of opensource code.

CODE/Sources/PennyBOT/Handlers/InteractionHandler.swift
gwynne: I'd be of a mind to just reject such names as invalid, honestly.

CODE/Sources/PennyAPI/Helps/HelpsHandler.swift
```diff
@@ -0,0 +1,82 @@
...
+    init(context: LambdaInitializationContext) async {
+        let awsClient = AWSClient(
+            httpClientProvider: .createNewWithEventLoopGroup(context.eventLoop)
+        )
+        self.awsClient = awsClient
+        self.helpsRepo = RepositoryFactory.makeHelpsRepository((awsClient, context.logger))
+        context.terminator.register(name: "Shutdown AWS", handler: { eventLoop in
+            eventLoop.makeFutureWithTask {
+                try await awsClient.shutdown()
+            }
```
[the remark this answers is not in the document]
MahdiBM: Benny recently mentioned this won't actually do anything other than maybe in tests ... so i might just remove it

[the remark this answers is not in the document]
MahdiBM: yeah i've removed these from other lambdas ... removed it form this lambda too

CODE/Sources/PennyBOT/Handlers/InteractionHandler.swift
MahdiBM: hmmm ... ok

.github/workflows/helpsLambdaUpload.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v3.0.0
+      - name: Build and zip
+        run: |
+          cd ./CODE 
+          ./scripts/build-and-package.sh HelpsLambda
+      - name: Remove build file
+        run: |
+          cd './CODE/.build/lambda/'
+          sudo rm -rf HelpsLambda
```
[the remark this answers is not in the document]
gwynne: Will look into this later.

PR #48, opened 2023-05-07
Report swift-proposal changes

MahdiBM:
Reports new swift proposals or those that have had a change of status.
Uses this url to get the data in JSON: https://download.swift.org/swift-evolution/proposals.json
The whole implementation is rather simple. Get the proposals through the link, check if there has been a change, then report.

* [x] Implemented
* [x] Tests

### DIFF
--- a/CODE/Sources/PennyBOT/+Array.swift
+++ b/CODE/Sources/PennyBOT/+Array.swift
@@ -1,0 +1,40 @@
+
+extension Array {
+    func divided(
+        _ isInLhs: (Element) async throws -> Bool
+    ) async rethrows -> (lhs: Array<Element>, rhs: Array<Element>) {
+        var lhs = ContiguousArray<Element>()
+        var rhs = ContiguousArray<Element>()
+
+        var iterator = self.makeIterator()
+
+        while let element = iterator.next() {
+            if try await isInLhs(element) {
+                lhs.append(element)
+            } else {
+                rhs.append(element)
+            }
+        }
+
+        return (Array(lhs), Array(rhs))
+    }
+
+    func divided(
+        _ isInLhs: (Element) throws -> Bool
+    ) rethrows -> (lhs: Array<Element>, rhs: Array<Element>) {
+        var lhs = ContiguousArray<Element>()
+        var rhs = ContiguousArray<Element>()
+
+        var iterator = self.makeIterator()
+
+        while let element = iterator.next() {
+            if try isInLhs(element) {
+                lhs.append(element)
+            } else {
+                rhs.append(element)
+            }
+        }
+
+        return (Array(lhs), Array(rhs))
+    }
+}
--- a/CODE/Sources/PennyBOT/+Array.swift
+++ b/CODE/Sources/PennyBOT/+Array.swift
@@ -23,6 +23,6 @@ extension Array {
     ) rethrows -> (lhs: ArraySlice<Element>, rhs: ArraySlice<Element>) {
         var copy = self
         let firstOfRhs = try copy.partition(by: isInLhs)
-        return (copy[copy.startIndex ..< firstOfRhs], copy[firstOfRhs ..< copy.endIndex])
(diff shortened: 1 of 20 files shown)


CODE/Sources/PennyBOT/+Array.swift
```diff
@@ -0,0 +1,40 @@
...
+        while let element = iterator.next() {
+            if try await isInLhs(element) {
+                lhs.append(element)
+            } else {
+                rhs.append(element)
+            }
+        }
+
+        return (Array(lhs), Array(rhs))
+    }
```
gwynne, line 20:
```suggestion
extension Array {
    func divided(
        _ isInLhs: (Element) async throws -> Bool
    ) async rethrows -> (lhs: [Element], rhs: [Element]) {
        var lhs = [Element]()
        var rhs = [Element]()
        lhs.reserveCapacity(self.count)
        rhs.reserveCapacity(self.count)

        for element in self {
            if try await isInLhs(element) {
                lhs.append(element)
            } else {
                rhs.append(element)
            }
        }
        return (lhs, rhs)
    }
```

CODE/Sources/PennyBOT/+Array.swift
```diff
@@ -0,0 +1,40 @@
...
+            if try isInLhs(element) {
+                lhs.append(element)
+            } else {
+                rhs.append(element)
+            }
+        }
+
+        return (Array(lhs), Array(rhs))
+    }
+}
```
gwynne, line 40:
```suggestion
    func divided(
        _ isInLhs: (Element) throws -> Bool
    ) rethrows -> (lhs: ArraySlice<Element>, rhs: ArraySlice<Element>) {
        var copy = self
        let firstOfRhs = try copy.partition(by: isInLhs)
        return (copy[copy.startIndex ..< firstOfRhs], copy[firstOfRhs ..< copy.endIndex])
    }
}
```

CODE/Sources/PennyBOT/ProposalsChecker.swift
```diff
@@ -0,0 +1,191 @@
+import Logging
+import DiscordModels
+import PennyModels
+import Foundation
+
+actor ProposalsChecker {
+    var previousProposals = [Proposal]()
+    var isFirstRound = true
```
gwynne, line 8:
```suggestion
```

CODE/Sources/PennyBOT/ProposalsChecker.swift
```diff
@@ -0,0 +1,191 @@
...
+
+    func check() async throws {
+        let proposals = try await self.proposalsService.get()
+
+        /// If it's the first round, we have nothing to compare to
+        if isFirstRound {
+            self.previousProposals = proposals
+            isFirstRound = false
+            return
+        }
```
gwynne, line 45:
```suggestion
        guard !self.previousProposals.isEmpty else {
            self.previousProposals = proposals
            return
        }
```

gwynne (approved):
I made a couple suggestions, but they're basically stylistic/idiom nits, nothing functional.

CODE/Sources/PennyBOT/+Array.swift
MahdiBM: thanks!

CODE/Sources/PennyBOT/+Array.swift
MahdiBM: thanks!

CODE/Sources/PennyBOT/ProposalsChecker.swift
MahdiBM: Thanks, i'll add this change with the next cleanup commit that's incoming

MahdiBM: Though i have something against `guard !thing else { }`, so i'll just use `if thing { }` 😅

CODE/Sources/PennyBOT/+Array.swift
MahdiBM: I think this one is causing some problems 👀 tests crashing now

PR #50, opened 2023-06-26
Across-boots data persistence

MahdiBM:


Sources/Penny/BotStateManager.swift
```diff
@@ -2,64 +2,123 @@ import DiscordBM
...
-    
+
+    func cancelIfCachePopulationTakesTooLong() {
+        Task {
+            try await Task.sleep(for: .seconds(15))
+            if !isCachePopulated {
+                canRespond = true
+                logger.error("No CacheStorage-population signal was done in-time")
+            }
+        }
```
gwynne: Traditionally one does this by keeping the `Task` handy, calling `cancel()` when populating is done, and having the task body be just the timeout case. This approach works but unnecessarily binds the task to the actor and can in theory suffer from a TOCTOU race (though it's quite extremely unlikely even for a race condition).

Sources/Penny/BotStateManager.swift
```diff
@@ -2,64 +2,123 @@ import DiscordBM
...
-              author.id.rawValue == Constants.botId,
-              message.content.hasPrefix(signal)
+              author.id.rawValue == Constants.botId
         else { return }
         guard let otherId = message.content.split(whereSeparator: \.isWhitespace).last else {
             logger.warning("Can't find id of the other Penny")
             return
         }
         if otherId == "\(self.id)" { return }
+        if message.content.hasPrefix(StateManagerSignal.shutdown.value) {
```
gwynne, line 74: Why `hasPrefix()` rather than `==`?

Sources/Penny/BotStateManager.swift
```diff
@@ -2,64 +2,123 @@ import DiscordBM
...
+            shutdown()
+        } else if message.content.hasPrefix(StateManagerSignal.didShutdown.value) {
+            populateCache()
+        } else {
+            logger.error("Unknown signal")
+            return
```
gwynne, line 80: Nit: This trick (also, say what the signal was...)
```suggestion
            return logger.error("Unknown signal", metadata: ["signal": .string(message.content)])
```

Sources/Penny/BotStateManager.swift
```diff
@@ -2,64 +2,123 @@ import DiscordBM
...
             logger.error("AWS has not yet shutdown this instance of Penny! Why?!")
         }
     }
-    
-    private func sendSignal() async {
+
+    private func populateCache() {
+        Task {
+            if isCachePopulated {
+                logger.warning("Received a did-shutdown signal but Cache is already populated")
```
gwynne, line 103: Why is this noteworthy?

Sources/Penny/BotStateManager.swift
```diff
@@ -2,64 +2,123 @@ import DiscordBM
...
-            payload: .init(content: signal + " \(self.id)")
+            payload: .init(
+                content: makeSignalMessage(text: signal.value, id: self.id)
+            )
         )
     }
-    
+
+    func makeSignalMessage(text: String, id: Int) -> String {
+        "\(text) \(id)"
```
gwynne, line 123: `guard !text.contains(where: { $0.isWhitespace }) else { .. }`, maybe (possibly as an `assert()`, if that would be programmer error)? Or is that something that can happen?

Sources/Penny/BotStateManager.swift
```diff
@@ -2,64 +2,123 @@ import DiscordBM
...
+    }
+
 #if DEBUG
     func _tests_reset() {
```
gwynne, line 127: Worth noting, generally having to do this means your unit tests aren't really "unit", since you're not testing actual code anymore.

Sources/Penny/BotStateManager.swift
```diff
@@ -68,5 +127,23 @@ actor BotStateManager {
...
+
+    var value: String {
+        switch self {
+        case .shutdown:
+            return "Hello the other Pennys 👋 you can retire now :)"
+        case .didShutdown:
+            return "I'm retired!"
+        }
+    }
+}
```
gwynne, line 153: Whilst I appreciate the personality of these messages (and you know I do! 🤣), perhaps this?
```suggestion
enum StateManagerSignal: String, CustomStringConvertible {
    case shutdown
    case didShutdown

    /// See ``Swift/CustomStringConvertible/description``.
    var description: String {
        switch self {
        case .shutdown:
            return "Hello the other Pennys 👋 you can retire now :)"
        case .didShutdown:
            return "I'm retired!"
        }
    }
}
```
(then use `.rawValue` where you were using `.value` elsewhere)

Sources/Penny/ProposalsChecker.swift
```diff
@@ -348,8 +353,8 @@ private extension Proposal.User {
     }
 }
 
-private struct QueuedProposal {
-    let uuid = UUID()
+struct QueuedProposal: Codable {
+    var uuid = UUID()
```
gwynne, line 357: Defaulted values and `Codable` are generally a bad mix, even if they're mutable.

Sources/Penny/Services/CacheService/CacheService.swift
```diff
@@ -0,0 +1,5 @@
+
```
gwynne, line 1:
```suggestion
```
Yep, pedantry about an extra newline 😆

Sources/Penny/Services/CacheService/CacheStorage.swift
```diff
@@ -0,0 +1,30 @@
...
+    static func makeFromCachedData() async -> CacheStorage {
+        var storage = CacheStorage()
+        storage.reactionCacheData = await ReactionCache.shared.toCacheStorageData()
+        storage.proposalsCheckerData = await ProposalsChecker.shared.toCacheStorageData()
+        return storage
+    }
+
+    func populateServices() async {
+        if let reactionCacheData = self.reactionCacheData {
+            await ReactionCache.shared.fromCacheStorageData(reactionCacheData)
```
gwynne: `await whatever.fromWhatever()` is not a particularly idiomatic formation. "From the cache storage data" _to where_? What is the verb, what action is being taken? A method with a name like this would typically be a factory method and/or return a value.

Sources/Penny/Services/CacheService/DefaultCacheService.swift
```diff
@@ -0,0 +1,51 @@
+import Logging
+import AsyncHTTPClient
+import SotoS3
+
+actor DefaultCachesService: CachesService {
+
```
gwynne, line 6:
```suggestion
```

Sources/Penny/Services/CacheService/DefaultCacheService.swift
```diff
@@ -0,0 +1,51 @@
...
+    let logger = Logger(label: "DefaultCachesService")
+
+    private init() { }
+
+    static let shared = DefaultCachesService()
+
+    func initialize(httpClient: HTTPClient) {
+        let awsClient = AWSClient(httpClientProvider: .shared(httpClient))
+        self.cacheRepo = .init(awsClient: awsClient, logger: self.logger)
+    }
```
gwynne, line 17: Ouch. Please dear Gods let [SE-0327](https://github.com/apple/swift-evolution/blob/main/proposals/0327-actor-initializers.md) actually show up in a Swift release soon...

Sources/Penny/Services/CacheService/DefaultCacheService.swift
```diff
@@ -0,0 +1,51 @@
...
+            self.delete()
+        } catch {
+            logger.report("Couldn't get CacheStorage", error: error)
+        }
+    }
+
+    /// Delete the object from the repository.
+    /// We don't care if it succeeds or not.
+    private func delete() {
+        Task {
```
gwynne, line 33: Not saying I have an alternative suggestion, to be sure, but as a point of general being the pedantic creature I am, I should point out that @fabianfett would be steaming at this much mixing of structured and unstructured concurrency almost as much as you might expect me to do at, say, one of Tanner's infamous `Codable` misuses 😂

Sources/Penny/Services/CacheService/S3CacheRepository.swift
```diff
@@ -0,0 +1,55 @@
...
+        self.logger = logger
+    }
+
+    func get() async throws -> CacheStorage {
+        let request = S3.GetObjectRequest(bucket: bucket, key: key)
+        let response = try await s3.getObject(request, logger: logger)
+
+        if let buffer = response.body?.asByteBuffer(), buffer.readableBytes != 0 {
+            do {
+                return try JSONDecoder().decode(CacheStorage.self, from: buffer)
```
gwynne: Initializing `JSONDecoder`s is (sometimes) expensive; consider making the decoder an instance property (a static would run into threading issues).

Stack.yml
```diff
@@ -402,6 +411,7 @@ Resources:
       PointInTimeRecoverySpecification:
         PointInTimeRecoveryEnabled: true
 
+
```
gwynne, line 414: Why extra newline?

Tests/PennyTests/GatewayProcessingTests.swift
```diff
@@ -43,11 +43,57 @@ class GatewayProcessingTests: XCTestCase {
...
     }
-    
+
+    func waitForStateManagerShutdownAndDidShutdownSignals() async {
+        /// Wait for the shutdown signal, then send a `didShutdown` signal.
+        /// in practice, the `didShutdown` signal is sent by another Penny that is online.
+        while let possibleSignal = await responseStorage.awaitResponse(
+            at: .createMessage(channelId: Constants.Channels.logs.id)
+        ).value as? Payloads.CreateMessage {
+            if possibleSignal.content?.contains(StateManagerSignal.shutdown.value) == true {
```
gwynne, line 57: `== true` is anti-idiomatically redundant; when dealing with the "oops `Bool?` isn't `Bool`" problem in conditionals I generally prefer `(foo.bar?.contains(...) ?? false)` unless doing so is excessively awkward in context. In a case like this, I'd probably go so far as to write `if let content = possibleSignal.content, content.contains(...)`

(Does it _really_ matter, especially in a test? Of course not, but I suspect it's hardly a revelation to anyone by now that I have a very particular idea of "good" Swift style 🙃.)

Tests/PennyTests/GatewayProcessingTests.swift
```diff
@@ -43,11 +43,57 @@ class GatewayProcessingTests: XCTestCase {
...
+        /// Wait to make sure the `BotStateManager` cache is already populated.
+        for idx in 1...100 {
+            try? await Task.sleep(for: .milliseconds(50))
+            if await BotStateManager.shared.isCachePopulated {
+                break
+            }
+            if idx == 100 {
+                XCTFail("BotStateManager cache was too late to populate")
+            }
+        }
```
gwynne, line 94: Ewww, polling loop... I don't have a better suggestion here short of Swift 5.9-specific features, though, beyond this simplification:

```suggestion
        for _ in 1...100 where !(await BotStateManager.shared.isCachePopulated) {
            try? await Task.sleep(for: .milliseconds(50))
        }
        XCTAssert(BotStateManager.shared.isCachePopulated, "BotStateManager cache was too late to populate")
```
(Note: I have not checked that you can actually do an await like that in a loop-conditional clause; if the compiler complains, put it in the body instead.)

gwynne (changes requested):
Several nits, questions, and notes. Also, should go without saying (but just in case), I fully expect you to go looking for other places you made a given issue (like not including enough data in an error message) on your own instead of needing me to add a comment to each one individually 😉.

gwynne (changes requested):
Much better, but one significant item I insist on putting my foot down about as a matter of not encouraging bad habits.

gwynne (approved):
LGTM, I think I've tormented you more than enough 🙂

PR #52, opened 2023-06-28
Penny report Github events using webhooks

### LINKED ISSUE #51: Report certain GitHub events in Discord channels
This includes:

- What [the release bot](https://github.com/vapor/release-bot) currently does, detecting PR merges and reacting to them in the PR and the Discord releases channel.
- Detect newly-opened issues and PRs in Vapor repos (maybe also in vapor-community repos?) and report them in Discord channels (TBD).

MahdiBM:
resolves #51


Lambdas/GHHooks/GHHooksHandler.swift
```diff
@@ -0,0 +1,81 @@
...
+        self.logger = context.logger
+    }
+
+    func handle(
+        _ event: APIGatewayV2Request,
+        context: LambdaContext
+    ) async throws -> APIGatewayV2Response {
+        try await verifyWebhookSignature(event: event)
+
+        #warning("fix return")
```
gwynne: 🤨

Sources/Extensions/APIGatewayExtensions.swift
```diff
@@ -7,13 +7,13 @@ private let jsonEncoder = JSONEncoder()
 
 extension APIGatewayV2Request {
     
-    public func decode<C: Codable>(as type: C.Type = C.self) throws -> C {
+    public func decode<D: Decodable>(as type: D.Type = D.self) throws -> D {
         guard let body = self.body,
               let dataBody = body.data(using: .utf8)
```
gwynne, line 12: Please prefer `Data(body.utf8View)` for getting `Data` from a `String` (and for going in the other direction, prefer `String(decoding: data, as: UTF8.self)`).

Lambdas/GHHooks/GithubEvent.swift
```diff
@@ -0,0 +1,281 @@
...
+final class DereferenceBox<C: Codable>: Codable {
+    var value: C
+
+    init(from decoder: any Decoder) throws {
+        self.value = try .init(from: decoder)
+    }
+
+    func encode(to encoder: any Encoder) throws {
+        try self.value.encode(to: encoder)
+    }
```
gwynne, line 280: This is an _**extremely**_ fine distinction that is completely lost on most uses (and users) of `Codable`, since it only makes a difference in very atypical cases, but those cases do actually come up now and then. The correct way to do this is:
```suggestion
    init(from decoder: any Decoder) throws {
        let container = try decoder.singleValueContainer()
        self.value = try container.decode(C.self)
    }

    func encode(to encoder: any Encoder) throws {
        var container = encoder.singleValueContainer()
        try container.encode(self.value)
    }
```

gwynne (changes requested):
The Codable thing is the only part I care about here.

gwynne (approved):
👍

Lambdas/GHHooks/GithubEvent.swift
MahdiBM: I actually had noticed the decoding difference, but I couldn't think of it when I was deciding which version to use, so just went with the more-consice version.

Still don't remember what it was.

PR #61, opened 2023-06-29
Use GH OpenAPI spec + swift-oapi-generator to generate GHHooks models

MahdiBM:
Uses Github OpenAPI spec + swift-openapi-generator for generating models for the GHHooks lambda.

The downside is the build time regression as the generator, at least as a plugin, seems not to be fast at all.
The upside is we won't need to make these models / api-endpoints in the future.


Package.swift
```diff
@@ -128,12 +136,14 @@ let package = Package(
...
+                .product(name: "OpenAPIGenerator", package: "swift-openapi-generator"),
+                .product(name: "OpenAPIRuntime", package: "swift-openapi-runtime"),
                 .product(name: "SotoSecretsManager", package: "soto"),
                 .product(name: "DiscordBM", package: "DiscordBM"),
                 .target(name: "Extensions"),
             ],
-            path: "./Lambdas/GHHooks",
-            swiftSettings: swiftSettings
+            path: "./Lambdas/GHHooks"
+//            , swiftSettings: swiftSettings
```
gwynne, line 146: Why?

MahdiBM: Because the generators generation don't use existential anys.

dannflor:
Why did you convert `html_url` to snake case but nothing else?

gwynne:
> Why did you convert `html_url` to snake case but nothing else?

Trying to match up the incorrect/inconsistent documentation of GitHub's API with what the API actually returns, I believe.

MahdiBM:
> Why did you convert html_url to snake case ...

Because that's what the OpenAPI generator generated.
Previously I used https://app.quicktype.io which automatically did snake -> camel conversions.

> but nothing else?

I'm not sure what do you mean by that ... as far as i can tell everything was changed to snake-case.

Kriangsak2536 (approved):

Kriangsak2536 (approved):

Lambdas/GHHooks/Models/Aliases.swift
```diff
@@ -0,0 +1,7 @@
+
```
gwynne, line 1: Always with the random blank lines even at the top of a file 🙂

Package.swift
gwynne: Good reason. Can I suggest 1) adding existential any-correctness to your draft PR on the generator, and 2) possibly moving the API definitions to a separate "just for the API definitions" target so we minimize the amount of code that loses out on the build setting(s)?

gwynne (approved):
Looks basically good to me, a couple of suggestions in comments but neither of them are really blocking so I leave it up to you whether to bother before merging.

Lambdas/GHHooks/Models/Aliases.swift
MahdiBM: I haven't found a general rule for when to put a blank line and when not, but generally i don't like starting definitions right at the first line.

Package.swift
MahdiBM: For the reference, will be solved by https://github.com/apple/swift-openapi-generator/pull/99.

PR #62, opened 2023-06-30
GH account link

ptoffy:
This aims to link discord user accounts to a GH account to be able to track contributions in the vapor ecosystem


Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,101 @@
...
+        guard let code = event.queryStringParameters?["code"] else {
+            return .init(
+                statusCode: .badRequest,
+                body: "Missing code query parameter"
+            )
+        }
+
+        logger.trace("Retrieving secrets")
+        let clientID = try await secretsRetriever.getSecret(arnEnvVarKey: "BOT_TOKEN_ARN")
+        let clientSecret = try await secretsRetriever.getSecret(arnEnvVarKey: "WH_SECRET_ARN")
```
MahdiBM, line 41: Client ID and Client secret are totally different from these.

@gwynne will need to create the oauth app, then add the Client Secret to AWS secrets manager, and also give us the Client ID.

in the stack file it'll look something like this: https://github.com/vapor/penny-bot/blob/main/deploy/penny-discord-bot-stack.yml#L394

More precisely:
```yaml
Environment:
  Variables:
    CLIENT_SECRET_ARN: arn:aws:secretsmanager:eu-west-1:177420307256:secret:prod/penny/penny-bot/FIXME_ADD_ARN_WHEN_GWYNNE_ADDS_TO_SECRETS_MANANGER
    CLIENT_ID: # just plain id like: 213230123910312
``` 

Then:
```swift
let clientSecret = try await secretsRetriever.getSecret(arnEnvVarKey: "CLIENT_SECRET_ARN")
guard let clientID = ProcessInfo.processInfo.environment["CLIENT_ID"] else {
    throw Errors.envVarNotFound(name: "CLIENT_ID")
}
```

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,101 @@
...
+        self.client = HTTPClient(eventLoopGroupProvider: .shared(context.eventLoop))
+        self.logger = context.logger
+        let awsClient = AWSClient(httpClientProvider: .shared(client))
+        self.secretsRetriever = SecretsRetriever(awsClient: awsClient, logger: logger)
+    }
+
+    func handle(
+        _ event: APIGatewayV2Request,
+        context: LambdaContext
+    ) async throws -> APIGatewayV2Response {
```
MahdiBM, line 29: This function should not be declared as `throws`.
Technically this _should be_ fine, but in practice, it won't be fine. If this function throws, AWS will just show a unhelpful 5xx to users.
Remove `throws`, and always return a response with a proper error description.

Sources/SharedServices/DynamoUserRepository.swift
```diff
@@ -39,7 +39,7 @@ struct DynamoUserRepository {
         let query = DynamoDB.QueryInput(
             expressionAttributeValues: [":v1": .s("DISCORD-\(id)")],
             indexName: discordIndex,
-            keyConditionExpression: "data1 = :v1",
+            keyConditionExpression: "discordID = :v1",
```
MahdiBM, line 42: You can't change `data1` here. This is a db query and there is no such column in the db as `discordID`.

Sources/SharedServices/DynamoUserRepository.swift
```diff
@@ -49,9 +49,9 @@ struct DynamoUserRepository {
     
     func getUser(github id: String) async throws -> DynamoUser? {
         let query = DynamoDB.QueryInput(
-            expressionAttributeValues: [":v1": .s("GITHUB-\(id)")],
+            expressionAttributeValues: [":v1": .s(id)],
             indexName: githubIndex,
-            keyConditionExpression: "data2 = :v1",
```
MahdiBM, line 54: You can't change `data2` here.

Sources/Models/DynamoDBUser.swift
```diff
@@ -3,8 +3,8 @@ import Foundation
 public struct DynamoDBUser: Sendable, Codable {
```
MahdiBM, line 3: You can't change this struct because it's used for decoding from database.
You can only change `DynamoUser`, not `DynamoDBUser`, until we migrate the db.

deploy/penny-discord-bot-stack.yml
```diff
@@ -417,6 +421,58 @@ Resources:
...
+      IntegrationMethod: POST
+      IntegrationType: AWS_PROXY
+      IntegrationUri: !GetAtt ghOAuthFunction.Arn
+      PayloadFormatVersion: '2.0'
+
+  ghOAuthRoutePost:
+    Type: AWS::ApiGatewayV2::Route
+    Properties:
+      ApiId: !Ref apiGateway
+      RouteKey: 'POST /gh-oauth'
```
MahdiBM, line 441: The redirect request is a 'GET', not 'POST'.

deploy/penny-discord-bot-stack.yml
```diff
@@ -417,6 +421,58 @@ Resources:
...
+  ghOAuthFunction:
+    Type: AWS::Lambda::Function
+    Properties:
+      Code:
+        S3Bucket: penny-discord-bot
+        S3Key: !Sub ${GHOAuthFilename}.zip
+      Environment:
+        Variables:
+          BOT_TOKEN_ARN: arn:aws:secretsmanager:eu-west-1:177420307256:secret:prod/penny/penny-bot/discord-bot-token-VmhBXH
+          WH_SECRET_ARN: arn:aws:secretsmanager:eu-west-1:177420307256:secret:prod/penny/penny-bot/github-webhook-secret-MASfKL
```
MahdiBM, line 453: Need to change these env vars when @gwynne has them ready, like i mentioned in another review comment.

deploy/penny-discord-bot-stack.yml
```diff
@@ -417,6 +421,58 @@ Resources:
...
+          LOG_LEVEL: trace
```
MahdiBM, line 454: You could keep log-level at trace for now till we are done debugging the lambda an we are sure it works OK.

MahdiBM (changes requested):
Some notes.
You also need to add more stuff to the stack file, here: https://github.com/vapor/penny-bot/blob/319414c34d9ec061de9c88f9c86c1d0e14491718/deploy/penny-discord-bot-stack.yml#L221
This is where permissions are. Basically copy-paste what other lambdas have, and also you'll need to add the secretsmanager arn like down there in the list, when @gwynne has it ready.

Sources/SharedServices/DynamoUserRepository.swift
ptoffy: I thought we wanted to rename these two fields? I know they are not renamed in the db yet, but it’s a starting point

MahdiBM: yeah believe me when i say i'm not satisfied with these names 😅, but we probably should let that be for another time/PR since @0xTim wants to rework the db too.

Lambdas/GHOAuth/OAuthLambda.swift
MahdiBM: We should probably redirect users back to our Discord server, but not sure how. For now you can just return a "we linked your account, get back to the discord server".

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -15,15 +15,42 @@ struct GHOAuthHandler: LambdaHandler {
...
         self.logger = context.logger
+
         let awsClient = AWSClient(httpClientProvider: .shared(client))
         self.secretsRetriever = SecretsRetriever(awsClient: awsClient, logger: logger)
+
+        let botToken = try await secretsRetriever.getSecret(arnEnvVarKey: "BOT_TOKEN_ARN")
+        self.discordClient = await DefaultDiscordClient(httpClient: client, token: botToken)
+    }
+
+    func handle(_ event: APIGatewayV2Request, context: LambdaContext) async throws -> APIGatewayV2Response {
```
MahdiBM, line 31: 😅 This will still now work well for us. You can keep the function as `throws`, but should try to actually not throw unless necessary:

So the difference between the `GHHooks` lambda and this lambda is that the other side of `GHHooks` lambda is Github who sends the events, but the other side of this lambda is an actual human.

With throwing anywhere, the lambda will show a 500 error to user with no description of what has happened. The lambda won't mention the error description etc...

That's why the `handle` function should not throw, unless absolutely necessary.
Throwing e.g. in the initializer when a secret or env var is not available, is fine. Because without a secret we can't continue our work, and not-finding a secret is an actual 500 error anyway since something in our code is not working.

But other than that you should try to not throw at all, unlike the GHHooks lambda (or if something needs to throw, use a do/catch instead)

Lambdas/GHOAuth/OAuthLambda.swift
MahdiBM: outdated

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -37,8 +64,10 @@ struct GHOAuthHandler: LambdaHandler {
         }
 
         logger.trace("Retrieving secrets")
-        let clientID = try await secretsRetriever.getSecret(arnEnvVarKey: "BOT_TOKEN_ARN")
-        let clientSecret = try await secretsRetriever.getSecret(arnEnvVarKey: "WH_SECRET_ARN")
+        let clientSecret = try await secretsRetriever.getSecret(arnEnvVarKey: "GH_CLIENT_SECRET_ARN")
+        guard let clientID = ProcessInfo.processInfo.environment["GH_CLIENT_ID"] else {
+            throw OAuthLambdaError.envVarNotFound(name: "GH_CLIENT_ID")
```
MahdiBM, line 69: These throws for example, are fine, because they are a 500 error anyway.

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -60,10 +89,7 @@ struct GHOAuthHandler: LambdaHandler {
         logger.trace("Got response: \(response.status)")
 
         guard response.status == .ok else {
-            return .init(
-                statusCode: .internalServerError,
-                body: "Unexpected response from GitHub: \(response.status)"
-            )
+            throw OAuthLambdaError.unexpectedResponse(statusCode: Int(response.status.code))
```
MahdiBM, line 92: This is not fine, return to the old just `return .init` statement.

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -83,19 +109,16 @@ struct GHOAuthHandler: LambdaHandler {
         let userResponse = try await client.execute(request, timeout: .seconds(30))
 
         guard userResponse.status == .ok else {
-            return .init(
-                statusCode: .internalServerError,
-                body: "Unexpected response from GitHub: \(userResponse.status)"
-            )
+            throw OAuthLambdaError.unexpectedResponse(statusCode: Int(userResponse.status.code))
```
MahdiBM, line 112: Use the old `return .init` statement.

Sources/SharedServices/DynamoUserRepository.swift
```diff
@@ -87,9 +87,14 @@ struct DynamoUserRepository {
 
     func linkGithub(with discordId: String, _ githubId: String) async throws -> String {
         // TODO: Wip
-        guard let discordUser = try await getUser(discord: discordId) else {
-            fatalError()
-        }
+        let query = DynamoDB.QueryInput(
+            expressionAttributeValues: [":v1": .s("DISCORD-\(discordId)")],
+            indexName: discordIndex,
+            keyConditionExpression: "discordID = :v1",
```
MahdiBM, line 93: still need to use `data1` unfortunately.
This query is basically equal to something like Fluent's 
```swift
DynamoDBUser.query(on: db)
.filter(\.$discordID == "DISCORD-\(discordId)")
.all()
```
The problem is, `DynamoDBUser` doesn't have `\.$discordID`, and for now only has `\.$data1`.

MahdiBM (changes requested):
Better, still some problems :)

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,124 @@
...
+        // https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps
+
+        logger.trace("Requesting GitHub access token")
+        var request = HTTPClientRequest(url: "https://github.com/login/oauth/access_token")
+        request.method = .POST
+        request.headers = [
+            "Accept": "application/json",
+            "Content-Type": "application/json"
+        ]
+        let requestBody = try JSONEncoder().encode([
```
MahdiBM, line 81: @gwynne said `JSONEncoder()` and `JSONDecoder()` might have initialization costs, so better to move all decoder/encoders to be a property of this struct so they are only initialized once, like:
```swift
let jsonDecoder = JSONDecoder()
let jsonEncoder = JSONEncoder()
```
then use those in functions where needed.

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,124 @@
...
+
+        let response = try await client.execute(request, timeout: .seconds(30))
+        logger.trace("Got response: \(response.status)")
+
+        guard response.status == .ok else {
+            throw OAuthLambdaError.unexpectedResponse(statusCode: Int(response.status.code))
+        }
+
+        let responseBody = try await response.body.collect(upTo: 1024 * 1024)
+        let accessToken = try JSONDecoder().decode(AccessTokenResponse.self, from: responseBody).accessToken
```
MahdiBM, line 96: same thing with encoder/decoders mentioned above.

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,124 @@
...
+        ]
+
+        let userResponse = try await client.execute(request, timeout: .seconds(30))
+
+        guard userResponse.status == .ok else {
+            throw OAuthLambdaError.unexpectedResponse(statusCode: Int(userResponse.status.code))
+        }
+
+        let userResponseBody = try await userResponse.body.collect(upTo: 1024 * 1024)
+        let id = try JSONDecoder().decode(User.self, from: userResponseBody).id
```
MahdiBM, line 116: same thing with encoder/decoders mentioned above.

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,124 @@
...
+
+        // https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user
+
+        logger.trace("Requesting GitHub user info")
+
+        request = HTTPClientRequest(url: "https://api.github.com/user")
+        request.method = .GET
+        request.headers = [
+            "Accept": "application/vnd.github+json",
+            "Authorization": "Bearer \(accessToken)"
```
MahdiBM, line 106: i guess you can add the `X-GitHub-Api-Version: 2022-11-28` header too like Github says, just to make sure they don't break our code.

deploy/penny-discord-bot-stack.yml
```diff
@@ -452,7 +452,8 @@ Resources:
       Environment:
         Variables:
           BOT_TOKEN_ARN: arn:aws:secretsmanager:eu-west-1:177420307256:secret:prod/penny/penny-bot/discord-bot-token-VmhBXH
-          # TODO: Add client id and secret
+          GH_CLIENT_SECRET_ARN: arn:aws:secretsmanager:eu-west-1:177420307256:secret:prod/penny/penny-bot/github-oauth-client-secret-hQSGeh
```
MahdiBM, line 455: Need to add the ARN to the permissions in the file too (they were around line 200 i think ... you already modified them once, a few commits ago)

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,136 @@
...
+            "Content-Type": "application/json"
+        ]
+        let requestBody = try jsonEncoder.encode([
+            "client_id": clientID,
+            "client_secret": clientSecret,
+            "code": code
+        ])
+        request.body = .bytes(requestBody)
+
+        let response = try await client.execute(request, timeout: .seconds(30))
```
MahdiBM, line 93: Throws from `client.execute`/`jsonDecoder`/`jsonEncoder` are still not fine.
You can probably put each one of these 2 "requests" in their own functions, and then do a single do/catch for those (instead of having to do a do/catch for very single `try`, which both is painful and also doesn't look good)

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -33,42 +33,39 @@ struct GHOAuthHandler: LambdaHandler {
...
-                        color: .red
-                    )])
-                ).guardSuccess()
-            } catch {
-                logger.error("DiscordClient logging error", metadata: [
-                    "error": "\(error)"
-                ])
-            }
-            throw error
+            return .init(statusCode: .badRequest, body: "Error getting access token: \(error)")
```
MahdiBM, line 48: Much better, but log the error using the logger, and just return a generic error message to users.
It'll both make sure we don't leak anything, and also a user can't do much with the error anyway/
```swift
logger.error("Error while getting the code", metadata: ["error": "\(error)"])
return .init(statusCode: .badRequest, body: "Error getting access token")
```

Sources/Penny/Handlers/InteractionHandler.swift
```diff
@@ -498,7 +498,10 @@ private extension InteractionHandler {
 
         switch subcommand {
         case GitHubSubCommand.link:
-            return "This command is still a WIP. Linking discordId: \(discordId) with GitHub account: \(String(describing: githubUsername))"
+            let clientID = try Constants.ghOAuthClientId.requireValue()
```
MahdiBM, line 501: just force-unwrap this. This a necessary environment value ... not-force-unwraping this necessary values will mostly just make the coding hard for us, than to make the code any more safe than it would be without the force unwrap.

Sources/Penny/Handlers/InteractionHandler.swift
```diff
@@ -486,17 +486,26 @@ private extension InteractionHandler {
...
+        if subcommand == .link, githubUsername == nil {
+            return "You need to provide a GitHub username to link your account."
+        }
+
+        switch subcommand {
+        case GitHubSubCommand.link:
+            let clientID = try Constants.ghOAuthClientId.requireValue()
+            let url = "https://github.com/login/oauth/authorize"
+            let githubLink = "\(url)?client_id=\(clientID)"
+            return "Use this to link your GitHub account to your Penny account: \(githubLink)"
```
MahdiBM, line 504: So we could be fancier. First of all, you need to add this extension to the end of the file (lines 1000-1200) where `protocol Response` and its extensions are:
```swift

extension Payloads.EditWebhookMessage: Response {
    func makeResponse(isEphemeral: Bool) -> Payloads.InteractionResponse {
        Logger(label: "Payloads.EditWebhookMessage.makeResponse").error(
            "This method is unimplemented and must not be called"
        )
        return .channelMessageWithSource(
            .init(content: "Oops, something went wrong")
        )
    }

    func makeEditPayload() -> Payloads.EditWebhookMessage {
        self
    }

    var isEditable: Bool { true }
}
```
Then you can return a payload with a button:
```swift
return Payloads.EditWebhookMessage(
                embeds: [.init(
                    description: "Click the link below to authorize Vapor",
                    color: .vaporPurple
                )],
                components: [[
                    .button(.init(label: "Authorize", url: githubLink))
                ]]
            )
```

Sources/Models/GHOAuthPayload.swift
```diff
@@ -0,0 +1,22 @@
+import DiscordBM
+import JWTKit
+
+/// This Payload will get sent along with the OAuth redirect to the GitHub OAuth page,
+/// specifically inside the `state` query parameter.
+/// This is used to verify the user's identity when they come back from GitHub 
+/// and that the request is not forged.
+public struct GHOAuthPayload: JWTPayload {
+    public let discordID: UserSnowflake
+    public let interactionID: InteractionSnowflake
```
MahdiBM, line 10: This was my bad, change this to `InteractionToken: String`:
```suggestion
    public let interactionToken: String
```

Sources/Models/GHOAuthPayload.swift
```diff
@@ -0,0 +1,22 @@
...
+    public let expiration: ExpirationClaim
+
+    public init(discordID: UserSnowflake, interactionID: InteractionSnowflake) {
+        self.discordID = discordID
+        self.interactionID = interactionID
```
MahdiBM, line 15: also here:
```suggestion
    public init(discordID: UserSnowflake, interactionToken: String) {
        self.discordID = discordID
        self. interactionToken = interactionToken
```

Sources/Penny/CommandsManager.swift
```diff
@@ -91,46 +91,41 @@ enum SlashCommand: String, CaseIterable {
...
-        case .slack: return [.init(
-            type: .string,
-            name: "id",
-            description: "Your Slack account ID",
-            required: true
-        )]
+        case .help:
+            return []
+        case .link:
+            return [.init(
```
MahdiBM, line 121: Now that we discussed everything on Discord, we know that we don't need the username option at all.

What happens is that Discord sends an `Interaction` payload our way, and the payload contains the Discord info of the user, including the Discord ID. so we can do something like `interaction.author.id` and get their ID.

The Interaction also contains a unique token, with which we can edit the interaction again from the Lambda.
So what happens is that we respond to the interaction in Discord part of penny and say "hey click on this link to link/button your GH". The link contains points to GitHub oauth-authorize endpoint, and the link contains an `state` parameter. 
The state parameter is the JWT token which you're supposed to create and sign using the **private** key. The `GHOAuthPayload` needs a `DiscordID` and `InteractionToken` which i mentioned both.

On the other side, the lambda gets the state parameter, verifies the state as you've done, and it'll have access to the same `GHOAuthPayload`.
After we have the OAuth access_token, the lambda uses the `discordID` in the `GHOAuthPayload` to do a penny-dynamo-db-table query and find the discord user, then update their GitHub ID and save the user to dynamo db. And linking is done :)

Sources/Penny/Constants.swift
```diff
@@ -11,6 +11,9 @@ enum Constants {
     static var botId: String! = env("BOT_APP_ID")
     static var loggingWebhookUrl: String! = env("LOGGING_WEBHOOK_URL")
     static var apiBaseUrl: String! = env("API_BASE_URL")
+    static var ghOAuthClientId: String! = env("GH_OAUTH_CLIENT_ID")
+    static var accountLinkOAuthPubKey: String! = env("ACCOUNT_LINKING_OAUTH_FLOW_PUB_KEY")
```
MahdiBM, line 15: public key not needed as explained

Sources/Penny/Handlers/InteractionHandler.swift
```diff
@@ -486,17 +516,38 @@ private extension InteractionHandler {
...
+        switch subcommand {
+        case GitHubSubCommand.link:
+            let clientID = Constants.ghOAuthClientId!
+            let jwt = GHOAuthPayload(
+                discordID: discordID, 
+                interactionID: event.id
+            )
+            let state = try ghOAuthJWTsigners.sign(jwt)
+            let url = "https://github.com/login/oauth/authorize?client_id=\(clientID)&state=\(state)"
+            return Payloads.EditWebhookMessage(
```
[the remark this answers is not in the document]
MahdiBM: I'm not 100% sure if a jwt token is already url-safe or not, need to check that.
if it isn't we'll need to url-encode the state.

Sources/SharedServices/DynamoUserRepository.swift
```diff
@@ -84,39 +84,19 @@ struct DynamoUserRepository {
...
+
+    func linkGithub(with discordId: String, _ githubId: String) async throws -> String {
+        // TODO: Wip
+        let query = DynamoDB.QueryInput(
+            expressionAttributeValues: [":v1": .s("DISCORD-\(discordId)")],
+            indexName: discordIndex,
+            keyConditionExpression: "data1 = :v1",
+            limit: 1,
+            tableName: self.tableName
+        )
```
MahdiBM, line 96: I know you haven't yet gotten to this, but you just need to add their github id and then probably call `updateUser()`

deploy/penny-discord-bot-stack.yml
```diff
@@ -435,6 +443,61 @@ Resources:
...
+      Code:
+        S3Bucket: penny-discord-bot
+        S3Key: !Sub ${GHOAuthFilename}.zip
+      Environment:
+        Variables:
+          BOT_TOKEN_ARN: arn:aws:secretsmanager:eu-west-1:177420307256:secret:prod/penny/penny-bot/discord-bot-token-VmhBXH
+          GH_CLIENT_SECRET_ARN: arn:aws:secretsmanager:eu-west-1:177420307256:secret:prod/penny/penny-bot/github-oauth-client-secret-hQSGeh
+          GH_CLIENT_ID: 5c544fca6b43f1bf31a3
+          ACCOUNT_LINKING_OAUTH_FLOW_PUB_KEY: LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KTUZrd0V3WUhLb1pJemowQ0FRWUlLb1pJemowREFRY0RRZ0FFZlR5bkhvSnEwNHhOaVN1cmRGK0JpcXVyRDh1NQp1RjdGU2V4OGFNalhXTVpreFgrZ0d3U3lwazBIMExvQ2g0LzFKK1Vhbkp4MzhWVDIwMVJpa2RVZ25BPT0KLS0tLS1FTkQgUFVCTElDIEtFWS0tLS0tCg==
+          ACCOUNT_LINKING_OAUTH_FLOW_PRIV_KEY: arn:aws:secretsmanager:eu-west-1:177420307256:secret:prod/penny/penny-bot/account-linking-oauth-flow-priv-key-49J2PL
```
MahdiBM, line 478: no private key needed

deploy/penny-discord-bot-stack.yml
```diff
@@ -244,11 +248,15 @@ Resources:
...
+                  - !Sub arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${GHOAuthFilename}-lambda
                   - !Sub arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:prod/penny/penny-bot/discord-bot-token-VmhBXH
                   - !Sub arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:prod/penny/penny-bot/trigger-gh-readme-workflow-token-cpZwMv
                   - !Sub arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:prod/penny/penny-bot/github-webhook-secret-MASfKL
+                  - !Sub arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:prod/penny/penny-bot/github-oauth-client-secret-hQSGeh
                   - !Sub arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:release-bot/github-access-token-pJOxYU
                   - !Sub arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:prod/penny/penny-bot/github-penny-app-private-key-vhi2WZ
+                  - !Sub arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:prod/penny/penny-bot/account-linking-oauth-flow-priv-key-49J2PL
```
MahdiBM, line 259: no private key needed

deploy/penny-discord-bot-task.json
```diff
@@ -28,6 +28,14 @@
                 {
                     "name": "BOT_APP_ID",
                     "value": "950695294906007573"
+                },
+                {
+                    "name": "GH_OAUTH_CLIENT_ID",
+                    "value": "5c544fca6b43f1bf31a3"
+                },
+                {
+                    "name": "ACCOUNT_LINKING_OAUTH_FLOW_PUB_KEY",
```
MahdiBM, line 37: no public key needed

deploy/penny-discord-bot-task.json
```diff
@@ -38,14 +46,18 @@
                 {
                     "valueFrom": "arn:aws:secretsmanager:eu-west-1:177420307256:secret:prod/penny/penny-bot/logs-webhook-url-4WnJwm",
                     "name": "LOGGING_WEBHOOK_URL"
+                },
+                {
+                    "valueFrom": "arn:aws:secretsmanager:eu-west-1:177420307256:secret:prod/penny/penny-bot/account-linking-oauth-flow-priv-key-49J2PL",
```
MahdiBM, line 51: good :)

MahdiBM (changes requested):
Nice work, although with some problems 😅
We're getting close!

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -59,12 +59,19 @@ struct GHOAuthHandler: LambdaHandler {
             return .init(statusCode: .badRequest, body: "Error getting user")
         }
 
+        do {
+            let jwt = try await verifyState(state: String(event.queryStringParameters?["state"] ?? ""))
+        } catch {
+            logger.error("Error verifying state: \(error)")
```
MahdiBM, line 65: log the state too please

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -78,19 +85,13 @@ struct GHOAuthHandler: LambdaHandler {
...
-        }
+        let signer = JWTSigner.es256(key: publicKey)
 
-        let signers = JWTSigners()
-        signers.use(.es256(key: privateKey))
-        signers.use(.es256(key: publicKey))
-
-        guard (try? signers.verify(state, as: GHOAuthPayload.self)) != nil else {
-            logger.trace("State parameter verification failed")
+        guard let payload = try? signer.verify(state, as: GHOAuthPayload.self) else {
```
MahdiBM, line 90: considering the `handle()` method catches this error, i think we're fine with the raw try and no guard

Sources/Penny/Handlers/InteractionHandler.swift
```diff
@@ -533,7 +523,11 @@ private extension InteractionHandler {
                 discordID: discordID, 
                 interactionID: event.id
             )
-            let state = try ghOAuthJWTsigners.sign(jwt)
+            guard let signer = makeJWTSigners() else {
+                logger.error("Failed to make JWT signer")
+                return oops
```
MahdiBM, line 528: good :)

MahdiBM (changes requested):
almost there :)

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -53,20 +57,27 @@ struct GHOAuthHandler: LambdaHandler {
...
             return .init(statusCode: .badRequest, body: "Error getting user")
         }
 
+        let jwt: GHOAuthPayload
+
         do {
-            let jwt = try await verifyState(state: String(event.queryStringParameters?["state"] ?? ""))
+            jwt = try await verifyState(state: String(event.queryStringParameters?["state"] ?? ""))
         } catch {
             logger.error("Error during state verification: error: \(error), state: \(event.queryStringParameters?["state"] ?? "")")
```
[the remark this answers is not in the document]
MahdiBM: optimally errors should be in metadata too, if you feel like it.
```
"error": "\(error)"
```

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -53,20 +57,27 @@ struct GHOAuthHandler: LambdaHandler {
...
             return .init(statusCode: .badRequest, body: "Error verifying state")
         }
 
-        // TODO: Link id to Discord user
+        do {
+            try await userService.linkUser(discordID: jwt.discordID.rawValue, githubID: userID.description)
```
MahdiBM, line 76: Swift team advices to use direct string interpolation instead of `.description`. I'm not sure why, maybe it's easier for the compiler to optimize that code:
```suggestion
try await userService.linkUser(discordID: jwt.discordID.rawValue, githubID: "\(userID)")
```

Sources/Models/DynamoDBUser.swift
```diff
@@ -22,4 +40,17 @@ public struct DynamoDBUser: Sendable, Codable {
         self.coinEntries = user.coinEntries
         self.createdAt = user.createdAt
     }
+
+    // Write a method that updates the user's data2 field to the githubID
+    public func updateGithubID(with githubID: String) -> DynamoDBUser {
```
MahdiBM, line 45: hmm the function seems like overkill for just changing the github id 🤔

Sources/SharedServices/DynamoUserRepository.swift
```diff
@@ -85,18 +85,33 @@ struct DynamoUserRepository {
     
     // MARK: - Link users
 
-    func linkGithub(with discordId: String, _ githubId: String) async throws -> String {
-        // TODO: Wip
+    func linkGithub(with discordID: String, _ githubID: String) async throws {
+        logger.trace("Linking Github ID \(githubID) with Discord ID \(discordID)")
+        enum Error: Swift.Error {
```
MahdiBM, line 90: Move the error to somewhere with higher level, like in the struct.

Sources/SharedServices/UserService.swift
```diff
@@ -96,6 +96,11 @@ public struct UserService {
     public func getUserWith(githubID id: String) async throws -> DynamoUser? {
         try await userRepo.getUser(github: id)
     }
+
+    /// Links an existing discord user to a github account.
+    public func linkUser(discordID: String, githubID: String) async throws {
+        try await userRepo.linkGithub(with: discordID, githubID)
```
MahdiBM, line 102: This whole thing seems too confusing, maybe try something like `linkGithubID(discordID: String, githubID: String)` then also change the userRepo function to something more readable like the same thing as this function.

MahdiBM (changes requested):
General nit-picky stuff, the implementation seems good 👌

MahdiBM (approved):

MahdiBM:
So far so good :)

MahdiBM:
I know the "request for review" thingy is something GitHub automatically does, and @ptoffy probably didn't mean it.
Just want to say @0xTim / @gwynne can wait a little bit more and i'll request their review later.

Sources/Models/GHOAuthPayload.swift
```diff
@@ -0,0 +1,22 @@
+import DiscordBM
+import JWTKit
+
+/// This Payload will get sent along with the OAuth redirect to the GitHub OAuth page,
+/// specifically inside the `state` query parameter.
+/// This is used to verify the user's identity when they come back from GitHub 
+/// and that the request is not forged.
+public struct GHOAuthPayload: JWTPayload {
```
MahdiBM, line 8: See what i said about `Models` in a `Package.swift` comment.

Sources/Penny/CommandsManager.swift
MahdiBM, line 121: See https://github.com/vapor/penny-bot/pull/62/files#r1265515491

Sources/Penny/Handlers/InteractionHandler.swift
```diff
@@ -27,11 +29,29 @@ struct InteractionHandler {
...
     }
+
+    private func makeJWTSigners() -> JWTSigner? {
+        let signer: JWTSigner
+
+        guard let privateKeyString = Constants.accountLinkOAuthPrivKey else {
+            fatalError("Missing ACCOUNT_LINKING_OAUTH_FLOW_PRIV_KEY env var")
+        }
+
+        guard let privateKey = try? ECDSAKey.private(pem: privateKeyString) else {
```
MahdiBM, line 46: It should be fine for the function to throw. Change this to `try`.
The command-handling methods will automatically send an "oops" if the try fails, IIRC.

Sources/Penny/Handlers/InteractionHandler.swift
```diff
@@ -27,11 +29,29 @@ struct InteractionHandler {
...
     typealias InteractionOption = Interaction.ApplicationCommand.Option
+
     
     init(event: Interaction) {
         self.event = event
         self.logger[metadataKey: "event"] = "\(event)"
     }
+
+    private func makeJWTSigners() -> JWTSigner? {
+        let signer: JWTSigner
```
MahdiBM, line 40: I think this is left-over from the early states of this method? You can combine it with what is in line 51

Sources/SharedServices/DynamoUserRepository.swift
```diff
@@ -49,7 +59,7 @@ struct DynamoUserRepository {
     
     func getUser(github id: String) async throws -> DynamoUser? {
```
MahdiBM, line 60: should just be `githubID`

Sources/SharedServices/DynamoUserRepository.swift
```diff
@@ -84,39 +94,31 @@ struct DynamoUserRepository {
...
-        // Return true if the merge was successful
-        abort()
-    }
-    
-    private func deleteAccount() async throws -> Bool {
-        // TODO: Implement
-        // Return true if the deletion was successful
-        abort()
+
+    func linkGithubID(to discordID: String, _ githubID: String) async throws {
```
MahdiBM, line 98: just do `linkGithubID(discordID: String, githubID: String)` so it's clear at call sites what is exactly happening and what the developer is supposed to pass to the function, and where.

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,171 @@
...
+            return .init(statusCode: .badRequest, body: "Error verifying state")
+        }
+
+        do {
+            try await userService.linkGithubID(to: jwt.discordID.rawValue, githubID: "\(userID)")
+        } catch {
+            logger.error("Error linking user with Discord ID \(jwt.discordID) and GitHub ID \(userID): \(error)")
+            return .init(statusCode: .badRequest, body: "Error linking user")
+        }
+
```
MahdiBM, line 83: Here, also update the Discord message using the `interactionToken` in the JWT:
```suggestion

        do {
            try await discordClient.updateOriginalInteractionResponse(
                token: jwt.interactionToken,
                payload: .init(
                    embeds: [.init(
                        description: """
                    Successfully linked your Discord account to GitHub account with ID: \(userID)
                    """,
                        color: .green
                    )]
                )
            ).guardSuccess()
        } catch {
            logger.warning("Received Discord error while updating interaction", metadata: [
                "error": "\(error)"
            ])
        }

```

Sources/SharedServices/DynamoUserRepository.swift
```diff
@@ -4,6 +4,16 @@ import Models
 import Extensions
 
 struct DynamoUserRepository {
+    enum Error: Swift.Error {
```
MahdiBM, line 7: Rename to `Errors`, conform to `CustomStringConvertible`.

Sources/SharedServices/DynamoUserRepository.swift
```diff
@@ -4,6 +4,16 @@ import Models
...
+        case discordUserNotFound(id: String)
+
+        var localizedDescription: String {
```
MahdiBM, line 10: Rename to `description`.

Realistically, i don't think you ever want to use `localizedDescription`.
What you usually want to use is a conformance to `CustomStringConvertible` followed by it's requirement, aka having `var description: String`. 
When you want a string representation of any type, Swift will have to resort to using `Mirror` to find that type's info at runtime, and generate a description for it (e.g. in string interpolation).
But when a type is conformed to `CustomStringConvertible`, Swift will be smart enough to use the `description` as the string representation, instead of using `Mirror`. This is good because using `Mirror` is known to be very slow.

Optionally, you might want to conform your Error types to `LocalizedError` too (need to import `Foundation`), and add a `var errorDescription: String?`. This is only useful for places that use Objective C runtime, like on macOS, iOS etc... . Overriding `errorDescription` will basically make it so you never see those unhelpful `The operation couldn’t be completed` errors, which is related to when an `Error` is converted to `NSError` and then you print it.

Lambdas/GHOAuth/OAuthLambda.swift
MahdiBM: Bonus points if you bother to change the `User` struct to also contain `let login: String`, which is basically the name of a GitHub account. Then mention the `login` in the Discord message, like `Successfully linked your Discord account to GitHub account with name: \(user.login)`.

You'll need to change the `getGHUserID` func to `getGHUser` and make it return the whole `User` struct so you can access the `user.login` for the Discord message.

MahdiBM (changes requested):
So these are _almost_ all suboptimalities i could find. Would be nice if you resolved these too.

Side Note: I personally try to follow the theme the project already has when trying to contribute, and don't try to force my own code style even if i think my code style is better. It's because having different code styles in a project is even worse than having just 1 singular bad code style, IMO.
Don't worry though i know you didn't mean anything special where you diverged from the general looks of the project (e.g. the functions name, errors names etc...). Just mentioning it as a general FYI 😅.

MahdiBM:
Final note (for now 😛): make the CI pass.
The `testLinkCommand()` is failing. If you feel like it, rename/repair it for the new GitHub command. Otherwise just get rid of it as the link command is renamed/removed.

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,196 @@
...
+            ])
+            return .init(statusCode: .badRequest, body: "Error verifying state")
+        }
+
+        do {
+            try await userService.linkGithubID(discordID: jwt.discordID.rawValue, githubID: "\(user.id)")
+        } catch {
+            logger.error("Error linking user to GitHub account", metadata: [
+                "discordID": .string(jwt.discordID.rawValue),
+                "githubID": .string("\(user.id)")
```
MahdiBM, line 92: both values conform to `CustomStringConvertible` so this can be simplified to:
```suggestion
                "discordID": .stringConvertible(jwt.discordID),
                "githubID": .stringConvertible(user.id)
```

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,196 @@
...
+            ])
+            return .init(statusCode: .badRequest, body: "Error linking user")
+        }
+
+        do {
+            try await discordClient.updateOriginalInteractionResponse(
+                token: jwt.interactionToken,
+                payload: .init(
+                    embeds: [.init(
+                        description: "Successfully linked your Discord account to GitHub account with username: \(user.id)",
```
MahdiBM, line 102:
```suggestion
                        description: "Successfully linked your Discord account to GitHub account with username: \(user.login)",
```

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,196 @@
...
+        return accessToken
+    }
+
+    func getGHUser(accessToken: String) async throws -> User {
+        logger.debug("Requesting GitHub user info")
+
+        // https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user
+
+        var request = HTTPClientRequest(url: "https://api.github.com/user")
+        request.method = .GET
```
MahdiBM, line 176: It's normal for requests to have `.GET` as the default, so i'd just remove this line.

Sources/Models/DynamoDBUser.swift
```diff
@@ -8,6 +8,24 @@ public struct DynamoDBUser: Sendable, Codable {
     public let amountOfCoins: Int?
     public let coinEntries: [CoinEntry]?
     public let createdAt: Date
+
+    public init(
```
MahdiBM, line 12: Seems to be unused. Remove.

Sources/Penny/CommandsManager.swift
```diff
@@ -91,46 +91,36 @@ enum SlashCommand: String, CaseIterable {
...
-            name: "id",
-            description: "Your Slack account ID",
-            required: true
-        )]
+        case .help:
+            return []
+        case .link:
+            return []
+        case .unlink:
+            return []
```
MahdiBM, line 123: you can just clamp them all into the same line:
```suggestion
        case .help, .link, .unlink:
            return []
```

MahdiBM (changes requested):
Thanks for going through all the requested changes 🙂.
Last round hopefully.
@gwynne please have a review if you want, then we can merge.

MahdiBM (approved):

Lambdas/GHOAuth/Models/User.swift
```diff
@@ -0,0 +1,4 @@
+struct User: Decodable {
```
gwynne, line 1:
```suggestion
struct User: Codable {
```
(If only because there's no reason _not_ to.)

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,194 @@
...
+            return .init(statusCode: .badRequest, body: "Missing code query parameter")
+        }
+
+        let accessToken: String
+
+        do {
+            accessToken = try await getGHAccessToken(code: code)
+        } catch {
+            logger.error("Error getting access token", metadata: [
+                "error": "\(error)"
```
gwynne, line 58:
```suggestion
                "error": "\(String(reflecting: error))"
```
I've been trying to get into the habit of always logging errors with their debug descriptions (i.e. with `String(reflecting:)`) rather than with their "default" descriptions; many errors provide much more useful info that way. (See also the debacle Fabian and I went through with making PostgresNIO's errors less info-leaking by default.)

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,194 @@
...
+            return .init(statusCode: .badRequest, body: "Error getting access token")
+        }
+
+        let user: User
+
+        do {
+            user = try await getGHUser(accessToken: accessToken)
+        } catch {
+            logger.error("Error getting user ID", metadata: [
+                "error": "\(error)"
```
gwynne, line 69:
```suggestion
                "error": "\(String(reflecting: error))"
```
And so on for other occurrences

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,194 @@
...
+        guard let publicKeyData = Data(base64Encoded: publicKeyString) else {
+            throw Errors.invalidPublicKey
+        }
+        let publicKey = try ECDSAKey.public(pem: publicKeyData)
+
+        let signer = JWTSigner.es256(key: publicKey)
+
+        let payload = try signer.verify(state, as: GHOAuthPayload.self)
+
+        return payload
```
gwynne, line 129:
```suggestion
        let jwtJsonEncoder = JSONEncoder(); jwtJsonEncoder.dateEncodingStrategy = .integerSecondsSince1970
        let jwtJsonDecoder = JSONDecoder(); jwtJsonDecoder.dateDecodingStrategy = .integerSecondsSince1970
        let signers = JWTSigners(defaultJSONEncoder: jwtJsonEncoder, defaultJSONDecoder: jwtJsonDecoder)
        signers.use(.es256(key: publicKey))
        return try signers.verify(state, as: GHOAuthPayload.self)
```
- Using a `JWTSigner` by itself is (very softly) deprecated in JWTKit. There aren't any warnings or even really documentation to that effect at this time, but it's true nonetheless. Going through `JWTSigners` is always recommended.
- Now that JWTKit's API supports it (which is a brand-new thing), we can actually configure the JSON coders appropriately.

I haven't included it in this suggestion, but the `JWTSigners` object should be created _once_ (including the loading of the key and such) during `init` and saved in a stored property rather than all of that work being done repeatedly here. (Yes I know that's of limited value given the nature of lambdas, but just the same.)

Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -0,0 +1,194 @@
...
+            "X-GitHub-Api-Version": "2022-11-28"
+        ]
+
+        let response = try await client.execute(request, timeout: .seconds(30))
+
+        guard response.status == .ok else {
+            throw Errors.badResponse(status: Int(response.status.code))
+        }
+
+        let userResponseBody = try await response.body.collect(upTo: 1024 * 1024)
```
gwynne, line 187: Where are these magic `upTo` numbers coming from in the various calls to `collect()`? Here we have 1MiB, a bit further up we have 4MiB, and in one of the other lambdas I've seen values like 64KiB used.

Sources/Penny/Handlers/InteractionHandler.swift
```diff
@@ -27,11 +29,22 @@ struct InteractionHandler {
...
     }
+
+    private func makeJWTSigners() throws -> JWTSigner? {
+        guard let privateKeyString = Constants.accountLinkOAuthPrivKey else {
+            fatalError("Missing ACCOUNT_LINKING_OAUTH_FLOW_PRIV_KEY env var")
+        }
+
+        let privateKey = try ECDSAKey.private(pem: privateKeyString)
+
+        return JWTSigner.es256(key: privateKey)
```
gwynne, line 46: Again, use `JWTSigners`. (See example and more details discussion earlier on.)

gwynne (changes requested):
Sorry to nitpick after all this PR has already been through 😕 My concerns are fairly minor, at least.

Lambdas/GHOAuth/OAuthLambda.swift
MahdiBM: Mostly just what I thought might be good when I wrote the code. Though not really supercalculated or something.

PR #63, opened 2023-07-04
Replace release bot

MahdiBM:
### Check list
- [x] Catch PR closed events where `merged` is `true` and detect the semver-patch/minor/major labels.
- [x] Create new releases.
  - [x] Determine the latest release through the API. Look for the `latest` tag. Don't just do `releases.last`.
- [x] Add comments on PRs.
- [x] Add auth headers and stuff to Github requests.
- [x] Send message to Discord.
- [x] Add GH-related secrets.
- [x] /latest-release doesn't give prereleases?
- [ ] Before deploy, consider it'll duplicate everything the release bot does (releases, comments etc..).
- [ ] Open up the release channel for Penny to send messages to.
  - Don't do it before merge or there will be the problem below:
- [x] Make sure Penny doesn't send thanks-responses to the release channel since it'll have send-messages perm.
- [x] Tests


Lambdas/GHHooks/Constants.swift
```diff
@@ -4,6 +4,7 @@ enum Constants {
     enum Channels: ChannelSnowflake {
         case logs = "1067060193982156880"
         case issueAndPRs = "1123702585006768228"
+        case release = "431926479752921098"
```
gwynne, line 7: Not a thing for this PR, of course, but as a matter of a _possible_ future TODO, does Discord's API allow for looking up these IDs versus just hardcoding them?

Lambdas/GHHooks/EventHandler/EventHandler.swift
```diff
@@ -0,0 +1,53 @@
...
+            )])
+        ).guardSuccess()
+    }
+}
+
+extension String {
+    func unicodesPrefix(_ maxLength: Int) -> String {
+        String(self.unicodeScalars.prefix(maxLength))
+    }
+}
```
gwynne, line 53: This can, of course, result in truncated display strings - a Unicode scalar is not even remotely a grapheme cluster. For example, take 👩‍👩‍👧‍👧, which is composed of `U+1F469 WOMAN` `U+200D ZERO WIDTH JOINER` `U+1F469 WOMAN` `U+200D ZERO WIDTH JOINER` `U+1F467 GIRL` `U+200D ZERO WIDTH JOINER` `U+1F467 GIRL`. Or even more fun, 👍 (`U+1F44D THUMBS UP`) versus 👍🏿(`U+1F44D THUMBS UP` `U+1F3FF EMOJI MODIFIER FITZPATRICK TYPE-6`). Breaking up strings at any granularity finer than extended grapheme cluster boundaries can have nontrivial consequences. Yes, even in the context where I see this method being used. Please take the prefix of the string itself.

Lambdas/GHHooks/EventHandler/Handlers/PRHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+            acknowledgment = "This patch was authored and released by @\(pr.user.login)."
+        } else {
+            acknowledgment = "This patch was authored by @\(pr.user.login) and released by @\(mergedBy.login)."
+        }
+
+        let release = try await makeNewRelease(version: version, acknowledgment: acknowledgment)
+
+        let ownerLogin = repo.owner.login
+        let repoName = repo.name
+        let url = "https://github.com/\(ownerLogin)/\(repoName)/releases/tag/\(release.tag_name)"
```
gwynne, line 101: Ugh. Please at least _consider_ constructing URLs via `URLComponents`, and not just in this one spot? Percent encoding matters, even if in these kinds of GitHub names it generally doesn't become an issue.

Lambdas/GHHooks/Models/+PR.Label.swift
```diff
@@ -0,0 +1,24 @@
...
+            KnownLabel(rawValue: $0.name)
+        }
+    }
+}
+
+enum KnownLabel: String {
+    case semVerMajor = "semver-major"
+    case semVerMinor = "semver-minor"
+    case semVerPatch = "semver-patch"
+    case release = "release"
```
gwynne, line 14: Suggest adding an additional case for the `no-release-needed` label (which admittedly doesn't exist on all repos yet but will when I eventually get around to it) - and with it a check to ensure that it isn't present in combination with one of the release labels (not to mention that there aren't multiple release labels on a single PR in general).

Lambdas/GHHooks/Models/SemVer.swift
```diff
@@ -0,0 +1,198 @@
+struct SemVer {
```
gwynne, line 1: \***cough**\* https://github.com/gwynne/swift-semver \***cough**\*...

gwynne (changes requested):
Per inline comments. You're gonna hate me for this one.

Lambdas/GHHooks/Constants.swift
MahdiBM: Yeah but then we'll be hadcoding channel names instead of ids.

Lambdas/GHHooks/EventHandler/EventHandler.swift
MahdiBM: I'm not going to act like I know all about graphemes ans stuff, but not using .unicodeScaalrs can easily result In bad-request from Discord as soon as there is an emoji in the String. .unicodeScalars has been working quite well while working with APIs for me. Not sure what else to do other than this.

Lambdas/GHHooks/EventHandler/Handlers/PRHandler.swift
MahdiBM: I copy pasted this from the old bot's code. It's just used for showing to users, not any actual API calls.
I could url encode them all but I don't think you're planning on changing vapor's org name to something weird, or using weird stuff in repo names 😅

Lambdas/GHHooks/Models/+PR.Label.swift
MahdiBM: Hmm it will already do no releases for any labels other than these so I don't think adding that label will do much 🤔

Lambdas/GHHooks/EventHandler/Handlers/PRHandler.swift
MahdiBM: Changed this url to use the release's `.html_url` since i noticed it has that.

For the record i do always consider url-encoding unless i forget 😅

Lambdas/GHHooks/Constants.swift
gwynne: Good point... I think I was conceptually reaching for "can this be configurable instead of forcing a rebuild and deploy to change them" ('cause we do that sooooooooooooooooooooo often and all 🤣)

MahdiBM: yeah it's suboptimal but not much we can do 😅
maybe moving all these to configuration files and making Penny read from those and ... but far from worth it.

Lambdas/GHHooks/EventHandler/EventHandler.swift
MahdiBM: I guess i could maybe find a way to not let a half-string scalar get in ... i might or might not give it a try 🤔

gwynne: Doing it "truly correctly" would be a rather intricate and messy task, but there's a simple cheat that's `O(1)` for the vast majority of cases:
```swift
    func unicodesPrefix(_ maxUnicodeScalars: Int) -> String {
        assert(maxUnicodeScalars >= 0, "Can request a negative maximum.")

        // Take a prefix of the string (i.e. a sequence of extended grapheme clusters) first.
        // Most often, of the time this will already be a result that suits our needs.
        var trimmed = self.prefix(maxUnicodeScalars)
        
        // If the result still has too many unicode scalars, there're one or more grapheme
        // clusters in the string. Keep dropping extended grapheme clusters off the end (which
        // with `String` is as easy as just removing the last `Character`) until we're within
        // bounds. Worst-case complexity is `O(n)`.
        while trimmed.unicodeScalars.count > maxUnicodeScalars {
            trimmed.removeLast()
        }
        
        return String(trimmed)
    }
```

Bonus chatter:

An "extended grapheme cluster" is any sequence of Unicode codepoints (in Swift parlance, a `UnicodeScalar`) which, when read in order from the beginning and applying Unicode's rules for combining characters, variation selectors, etc., yields a single "printable" glyph (where "printable" is a rather flexible term including such visually questionable esoterica as non-breaking spaces, LTR/RTL selectors, and so forth). Swift calls these `Character`s.

A "grapheme cluster" (formally, a "legacy grapheme cluster") is the same thing, except using the legacy (Unicode 3.2 or so) combining rules. These exist now only as a historical curiosity, not unlike the original FSS-UTF 6-byte version of UTF-8; you'll never have to worry about the difference.

Lambdas/GHHooks/EventHandler/Handlers/PRHandler.swift
gwynne: Always best to use what the API provides when it does so, definitely ❤️

Lambdas/GHHooks/Models/+PR.Label.swift
gwynne: It'll do something if you use it to do the conflict check 🙂. Which actually is something worth having, trust me 😰. Ideally, we'd eventually have it so that merging a PR that has neither semver _nor_ no-release labels would result in the person who did the merge getting a warning notification (which we can't do right now since not all repos have the label).

Lambdas/GHHooks/Constants.swift
gwynne: Yeah, agreed.

Lambdas/GHHooks/Models/SemVer.swift
MahdiBM: Done

gwynne (approved):
Okay, LGTM so far 🙂

Lambdas/GHHooks/Models/+PR.Label.swift
MahdiBM: not sure if i you-noticed/i-mentioned but added that label too.

MahdiBM:
There is a known warning in `GHLazyMiddleware` which will be solved by https://github.com/apple/swift-openapi-runtime/pull/22.

Lambdas/GHHooks/Models/+SemanticVersion.swift
```diff
@@ -0,0 +1,111 @@
...
+                version.major += 1
+            }
+        } else {
+            guard version.prereleaseIdentifiers.count < 4 else {
+                /// Shouldn't have more than 3 identifiers, otherwise unexpected by this code.
+                return nil
+            }
+            /// At this point we are guaranteed 1/2/3 `prereleaseIdentifiers`.
+
+            guard !version.prereleaseIdentifiers[0].allSatisfy(\.isNumber) else {
```
gwynne, line 35:
```suggestion
            guard version.prereleaseIdentifiers[0].first?.isLetter ?? false else {
```
Much less expensive check that will accept both (for example) "alpha" and "a1pha" while rejecting "123" and "1alpha"(which the `allSatisfy()` version accepts).

Lambdas/GHHooks/Models/+SemanticVersion.swift
```diff
@@ -0,0 +1,111 @@
...
+                /// Identifiers should be like `["alpha", "1"]`.
+                /// First identifier should not be a number.
+                return nil
+            }
+
+            guard version.prereleaseIdentifiers[1...]
+                .allSatisfy({ $0.allSatisfy(\.isNumber) }) else {
```
gwynne, line 42:
```suggestion
            guard version.prereleaseIdentifiers[1...].allSatisfy({ UInt($0) != nil }) else {
```
(The "allSatisfy is number" check would accept absurdly long numbers and empty strings, for example.)

Lambdas/GHHooks/Models/+SemanticVersion.swift
```diff
@@ -0,0 +1,111 @@
...
+            }
+        } else {
+            guard version.prereleaseIdentifiers.count < 4 else {
+                /// Shouldn't have more than 3 identifiers, otherwise unexpected by this code.
+                return nil
+            }
+            /// At this point we are guaranteed 1/2/3 `prereleaseIdentifiers`.
+
+            guard !version.prereleaseIdentifiers[0].allSatisfy(\.isNumber) else {
+                /// Identifiers should be like `["alpha", "1"]`.
```
gwynne, line 36: Might be worth defining an `enum ReleaseStage { case alpha, beta, rc }` type and checking for `ReleaseStage(rawValue: ...)` instead, come to think of it.

Lambdas/GHHooks/Models/+SemanticVersion.swift
```diff
@@ -0,0 +1,111 @@
...
+                let lastIndex = version.prereleaseIdentifiers.count - 1
+                if lastIndex == 1 {
+                    /// Doesn't have any minor identifiers. We add it.
+                    version.prereleaseIdentifiers.append("1")
+                } else {
+                    /// Already checked not-nil, but still trying to be safe.
+                    guard let number = Int(version.prereleaseIdentifiers[lastIndex])
+                    else { return nil }
+                    version.prereleaseIdentifiers[lastIndex] = "\(number + 1)"
+                }
```
gwynne, line 78:
```suggestion
                if version.prereleaseIdentifiers.count == 2 {
                    /// Doesn't have any minor identifiers. We add it.
                    version.prereleaseIdentifiers.append("1")
                } else {
                    guard let prev = Int(version.prereleaseIdentifiers.removeLast())
                    else { return nil }
                    version.prereleaseIdentifiers.append("\(prev + 1)")
                }
```

Sources/Penny/+String.swift
```diff
@@ -2,8 +2,22 @@ import Foundation
 import Models
 
 extension String {
```
gwynne, line 4: I get why this is duplicated here, but it would be very preferable for it not to be. (I realize that implies adding a separate utilities target to the package, though; I don't consider it a blocker for this PR.)

Sources/Penny/+String.swift
```diff
@@ -2,8 +2,22 @@ import Foundation
...
-    func unicodesPrefix(_ maxLength: Int) -> String {
-        String(self.unicodeScalars.prefix(maxLength))
+    func unicodesPrefix(_ maxUnicodeScalars: Int) -> String {
+        assert(maxUnicodeScalars >= 0, "Can request a negative maximum.")
```
gwynne, line 6: As mentioned before 😉

Tests/Fake/TestData.swift
```diff
@@ -65,4 +65,19 @@ public enum TestData {
...
         }
     }
+
+    private static let ghRestOperations: [String: Any] = {
+        let data = resource(named: "ghRestOperations.json")
+        let object = try! JSONSerialization.jsonObject(with: data, options: [])
+        return object as! [String: Any]
+    }()
+
+    /// Probably could be more efficient than encoding then decoding again?!
```
gwynne, line 75: Not unless you encode each individual operation as a data blob in the original JSON, which defeats the purpose of making it so conveniently structured. Not a big deal. I might recommend you do the re-serialize step as part of the "loading" step above, though, so that you end up with `[String: Data]`; that way you shift the all of the overhead into the setup stage rather than having it affect each test's execution time (normally you'd _want_ that kind of amortization, of course, but for tests an accurate performance measurement (even in a debug build) is generally more useful).

Tests/PennyTests/GHHooksTests.swift
```diff
@@ -15,45 +19,118 @@ class GHHooksTests: XCTestCase {
...
                 line: line
             ).value
-            if !expectNoResponse {
-                XCTAssertNotNil(response as? Payloads.CreateMessage, line: line)
+            if expectResponse {
+                XCTAssertTrue(
+                    type(of: response) == Payloads.CreateMessage.self,
+                    "'\(type(of: response))' is not equal to 'Payloads.CreateMessage'",
+                    line: line
+                )
```
gwynne, line 126:
```suggestion
                XCTAssertEqual(
                    "\(type(of: response)), "\(Payloads.CreateMessage.self),
                    line: line
                )
```

deploy/penny-discord-bot-stack.yml
```diff
@@ -511,8 +513,4 @@ Outputs:
     Value: !GetAtt addCoinFunction.Arn
 
   dynamoDBArn:
-    Value: !GetAtt dynamoUserTable.Arn
-
-  
-
-
+    Value: !GetAtt dynamoUserTable.Arn
```
gwynne, line 516:
```suggestion
    Value: !GetAtt dynamoUserTable.Arn

```
(note in case GH doesn't make it obvious: get rid of the "missing newline at end of file" symbol)

Tests/PennyTests/GHHooksTests.swift
gwynne, line 1: Please add a test (doesn't have to be in this particular file - in fact, it probably shouldn't be) that asserts that `unicodesPrefix(_:)` behaves correctly for strings consisting of multiple extended grapheme clusters.

Example strings:
- `"Hello, world! 👍🏾"` for max lengths 13-17 inclusive. Correct results will never contain `👍`.
- `"👍🏿👍🏾👍🏽👍🏼👍🏻👍"` for max lengths 1-15 inclusive. Correct results will contain `👍` only for lengths 14 and 15 , and for length 1 should be an empty string.
- `"👩‍👩‍👧‍👦👨‍👨‍👧‍👦"` for max lengths 1-15 inclusive. Correct results will always either be empty (lengths 1-6) or contain exactly one (lengths 7-13) or both (lengths 14-15) of the original emoji, never any others.

(I have not done any particular checking that my numbers here are correct, I'm just doing quick counting based on the number of codepoints Character Viewer says are in each emoji. If the results don't match up, look at the actual outputs displayed and make sure my numbers aren't just off before looking for a problem in the implementation. (But don't necessarily assume the implementation isn't wrong either!))

gwynne (changes requested):
Some of these are tiny nits, others are serious concerns. At least two are actual logic errors, and there's a major threading issue to boot. (At least, assuming I didn't misread anything or do the logic in my head wrong, we all know that happens a lot 😆)

gwynne:
@MahdiBM Watch out for Github's tendency to hide comments behind "5 hidden conversations" etc. links on the Conversation page of PRs; I miss comments that way more often than I care to admit 🤦‍♀️.

Lambdas/GHHooks/GHLazyMiddleware.swift
```diff
@@ -0,0 +1,113 @@
...
+        defer { isLoading = false }
+
+        let githubToken = try await secretsRetriever.getSecret(arnEnvVarKey: "GH_TOKEN_ARN")
+        self.githubToken = Secret(githubToken)
+
+        for waiter in self.loadWaiters {
+            waiter.resume()
+        }
+        self.loadWaiters.removeAll()
+    }
```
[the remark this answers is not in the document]
MahdiBM: The reason i went through the trouble of a "manual lock" was that i had heard somethings about different threads not working well with a lock .... i'm still not sure but i guess an actor works on just 1 thread anyway so we should be good?!

I think the manual lock i did would work fine too (the locking mechanism isn't even needed yet based on the fact that the whole lambda's execution is sequential, but i wanted to be safe for the future)

I'll move the locking mechanism to the `SecretsRetriever` because i have more plans for it ... I could make the `discordClient` be created lazily ... it matters because `discordClient` needs the bot token secret and getting each secret costs $$$.

Lambdas/GHHooks/Models/+SemanticVersion.swift
MahdiBM: It was like that but i didn't see a value in keeping the enum 🤔

MahdiBM: I guess i'll do it if you say so ... i don't really have a preference

deploy/penny-discord-bot-stack.yml
MahdiBM: I'm still not sure what you exactly meant but i think you meant to remove the last extra line?! i did that.

Sources/Penny/+String.swift
MahdiBM: yeah but i don't want to make a whole new target just because of 1 extension 😅 maybe when there are a few of them

Lambdas/GHHooks/GHLazyMiddleware.swift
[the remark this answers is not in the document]
MahdiBM: ah so i didn't read the change originally ...

I don't think the locking mechanism does anything special ... i think the simpler locking mechanism i had is safe (i've tried in a lot of places, other than asking people about it .... I'm still not 100% sure it's safe, but i think it is)

deploy/penny-discord-bot-stack.yml
MahdiBM:
```swift
await withCheckedContinuation { continuation in
    self.stateLock.withLockVoid { // prevent concurrent mutation
        self.loadWaiters.append(continuation)
    }
}
```
specifically here, i don't think any "concurrent mutations" can happen. 
I know there is a `await` there which means there is a suspension point, but i think the suspension point is triggered after the whole closure is executed, and is only used to wait for resumption of the continuation. A month or two ago I looked into the implementation of it but not through the right place, so didn't come to a proper conclusion.

Lambdas/GHHooks/GHLazyMiddleware.swift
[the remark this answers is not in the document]
MahdiBM:
```swift
await withCheckedContinuation { continuation in
    self.stateLock.withLockVoid { // prevent concurrent mutation
        self.loadWaiters.append(continuation)
    }
}
```
specifically here, i don't think any "concurrent mutations" can happen. 
I know there is a `await` there which means there is a suspension point, but i think the suspension point is triggered after the whole closure is executed, and is only used to wait for resumption of the continuation. A month or two ago I looked into the implementation of it but not through the right place, so didn't come to a proper conclusion.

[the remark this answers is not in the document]
MahdiBM:
```swift
self.stateLock.withLockVoid {
    self.isLoading = false
    while let waiter = self.loadWaiters.popLast() { // empties the array
        waiter.resume()
    }
    self.githubToken = try loadResult.get()
}
```
Here also ... there is no suspension point anyway so i don't think the lock does anything again.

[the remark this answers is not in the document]
MahdiBM: also the problem with `popLast()` is that it pops the last one  ... not the first one. There is also no `popFirst()`.
Not a big deal but i rather keep the order.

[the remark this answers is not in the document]
gwynne: As we discovered on Discord, this entire pile of nonsense I started was an utterly invalid thoroughly humiliating whopping error on my part 😅 That being said, just for the point of record - the waiter list by definition does not have a well-defined order, and even if it did, you wouldn't want it to. No matter how you consume it, by unblocking waiters in _any_ given order trashes the priority management semantics inherent to tasks and the lower-level locks they rely on (though moreso on macOS than Linux by a tremendous margin - Linux doesn't have priority-donating unfair locks, at least not in the sense of `os_unfair_lock_t`'s semantics). It's not even remotely worth caring about here, much like anything else I've said in this conversation thread 🤣, but it does mean that even if you were needing to deal with locking, the ordering of the waiters can never be well-defined without explicit serialization (which defeats the point) and resuming them in _any_ given order is always arguably wrong anyway, so random or inverse is as good as anything else. (Countdown until I figure out that this explanation is also irrelevant or based on a fundamental misreading: 10... 5... 8... -∞... π... TREE(TREE(TREE(pow(2, 128)))), √-1... 😂)

Lambdas/GHHooks/Models/+SemanticVersion.swift
gwynne: I do prefer it, though I'm too annoyed at myself to try to explain why just now 🤣

Sources/Penny/+String.swift
gwynne: Like I said, not a blocker here 🙂

deploy/penny-discord-bot-stack.yml
gwynne: What I meant was "put the one of the trailing newlines back" 😰 See screenshot:
Screenshot 2023-07-09 at 23 54 10

Getting rid of all the _extra_ trailing newlines, definitely a good thing, I heartily approve - but you left no newline after `Arn` at the end of the last line. Git considers that bad form (and I happen to agree).

Lambdas/GHHooks/SecretsRetriever.swift
```diff
@@ -23,6 +41,30 @@ struct SecretsRetriever {
         guard let secret = secret.secretString else {
             throw Errors.secretNotFound(arn: arn)
         }
-        return secret
+        return Secret(secret)
+    }
+
+    private func withLock<T>(block: () async throws -> T) async rethrows -> T {
+        if isLoading {
```
gwynne, line 48: To no one's surprise more than my own, I think one of my original suggestions here was actually still correct:
```suggestion
        while isLoading {
```
As per https://www.wwdcnotes.com/notes/wwdc21/10133/, when you cross an `await` in an `actor`, you must re-verify mutable state.

Lambdas/GHHooks/SecretsRetriever.swift
```diff
@@ -23,6 +41,30 @@ struct SecretsRetriever {
...
+
+        isLoading = true
+        defer { isLoading = false }
+
+        let value = try await block()
+
+        for waiter in self.loadWaiters {
+            waiter.resume()
+        }
+        self.loadWaiters.removeAll()
```
gwynne, line 62:
```suggestion
    isLoading = false
    self.loadWaiters.popLast()?.resume()
```
If you resume multiple waiters, you'll just wake them all up just to loop all but one of them back into another round of continuation awaits. By waking up only one, you get rid of a bunch of completely pointless contention on the Concurrency runtime's internal locks. It also avoids the possible race condition where two or more waiters get past the `isLoading` check before it gets set to `true` again.

But you do need to de-flag `isLoading` first, otherwise you'll very likely deadlock (`isLoading` will still be true, so it'll loop again, but there'll be no active loads to trigger another resume). You should remove the `defer` above too, un-flagging it twice can cause a different kind of TOCTOU race.

Lambdas/GHHooks/SecretsRetriever.swift
```diff
@@ -23,6 +41,30 @@ struct SecretsRetriever {
...
+        if isLoading {
+            await withCheckedContinuation { continuation in
+                loadWaiters.append(continuation)
+            }
+        }
+
+        isLoading = true
+        defer { isLoading = false }
+
+        let value = try await block()
```
gwynne, line 57: You allow a thrown error from here to exit the method without resuming any waiters, that'll deadlock anything that ended up in a wait during the failed load.

Tests/PennyTests/GHHooksTests.swift
```diff
@@ -15,46 +19,173 @@ class GHHooksTests: XCTestCase {
...
+        do {
+            let scalars_11 = "👍🏿👍🏾👍🏽👍🏼👍🏻👍"
+            let scalars_10 = "👍🏿👍🏾👍🏽👍🏼👍🏻"
+            let scalars_8 = "👍🏿👍🏾👍🏽👍🏼"
+            XCTAssertEqual(scalars_11.unicodesPrefix(12), scalars_11)
+            XCTAssertEqual(scalars_11.unicodesPrefix(11), scalars_11)
+            XCTAssertEqual(scalars_11.unicodesPrefix(10), scalars_10)
+            XCTAssertEqual(scalars_11.unicodesPrefix(9), scalars_8)
+            XCTAssertEqual(scalars_11.unicodesPrefix(8), scalars_8)
+        }
```
gwynne, line 47: Can you take this down one more grapheme (add lengths 7 and 6)? Just for safety's sake. I agree there's no need to test all the way down to 1.

gwynne (changes requested):
Almost there, I think 😰 Sorry that I ended up worrying about threading stuff _anyway_ after all that - I'm considerably less confident of my conclusions this time, but it nonetheless seems correct to me after careful reading of the linked info. Please do feel free to challenge my reasoning if you see any (more) mistakes 😅

Lambdas/GHHooks/GHLazyMiddleware.swift
```diff
@@ -0,0 +1,88 @@
...
+actor GHLazyMiddleware: ClientMiddleware {
+
+    private let secretsRetriever: SecretsRetriever
+    private let logger: Logger
+
+    private var idGenerator = 0
+
+    private var isLoading = false
+    private var loadWaiters: [CheckedContinuation<Void, Never>] = []
+
```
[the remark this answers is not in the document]
MahdiBM: aahh yes

gwynne (approved):
Okay, finally LGTM 😅 Excellent work!

PR #65, opened 2023-07-12
GHHooks lambda filter out html markdown blocks when posting to Discord

MahdiBM:
Use swift-markdown in GHHooks lambda to filter out html markdown blocks when posting to Discord.

Look at the added test to see what i mean.

### DIFF
--- a/Lambdas/GHHooks/Models/+Document.swift
+++ b/Lambdas/GHHooks/Models/+Document.swift
@@ -1,0 +1,10 @@
+import Markdown
+
+extension Document {
+    func filterOutChildren<T>(ofType: T.Type) -> Document where T: BlockMarkup {
+        let children = self.blockChildren.filter { markup in
+            type(of: markup) != ofType.self
+        }
+        return Document(children)
+    }
+}
--- a/Lambdas/GHHooks/Models/+Document.swift
+++ b/Lambdas/GHHooks/Models/+Document.swift
@@ -1,10 +1,7 @@
 import Markdown
 
 extension Document {
-    func filterOutChildren<T>(ofType: T.Type) -> Document where T: BlockMarkup {
-        let children = self.blockChildren.filter { markup in
-            type(of: markup) != ofType.self
-        }
-        return Document(children)
+    func filterOutChildren(ofType type: (some BlockMarkup).Type) -> Document {
+        Document(self.blockChildren.filter { type(of: $0) != type }
     }
 }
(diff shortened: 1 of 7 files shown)


Lambdas/GHHooks/Models/+Document.swift
```diff
@@ -0,0 +1,10 @@
+import Markdown
+
+extension Document {
+    func filterOutChildren<T>(ofType: T.Type) -> Document where T: BlockMarkup {
+        let children = self.blockChildren.filter { markup in
+            type(of: markup) != ofType.self
+        }
+        return Document(children)
+    }
```
gwynne, line 9:
```suggestion
    func filterOutChildren(ofType type: (some BlockMarkup).Type) -> Document {
        Document(self.blockChildren.filter { type(of: $0) != type }
    }
```
Might as well take advantage of all that fancy new stuff 5.7 enabled for generics 🙂.

Bonus chatter:
If I were writing for 5.6, or for any other reason couldn't simplify it with an opaque type, I'd write it like this:
```swift
    func filterOutChildren<T: BlockMarkup>(ofType _: T.Type) -> Document {
        Document(self.blockChildren.filter { type(of: $0) != T.self })
    }
```
According to the general rule that a non-opaque metatype parameter should always be referred to by its generic argument name, not the parameter's local name, whenever possible. (However, worth noting you _did_ correctly use an inequality comparison between the types. A common error in this kind of logic is to use a conformance check, e.g. `!(type(of: $0) is T.Type)`. (Apologies if the following is already known to you, I'm pontificating as usual 😆) The conformance test would happen to also have the desired result in this particular use case, but that's a syntactic quirk; a conformance test between metatypes checks "can type A be treated as type B" (which gets really fun when `Any.Type` gets involved), not "are type A and type B the same type".

Package.swift
```diff
@@ -49,6 +49,7 @@ let package = Package(
         .package(url: "https://github.com/DiscordBM/DiscordBM.git", branch: "main"),
         .package(url: "https://github.com/DiscordBM/DiscordLogger.git", from: "1.0.0-rc.1"),
         /// Not-released area:
+        .package(url: "https://github.com/apple/swift-markdown.git", .branch("main")),
```
gwynne, line 52: Give yourself a 10-🪙 (stupid Unicode definition of "coin"...) bonus for using SwiftMarkdown ❤️❤️❤️

gwynne (approved):
One style nit, which you can ignore if you feel like it. Very nice and neat! ❤️

PR #68, opened 2023-07-14
Use dedicated GitHub app

MahdiBM:


Lambdas/GHHooks/Extensions/Embed+Equtable.swift
```diff
@@ -42,13 +46,30 @@ extension Embed.Author: Equatable {
...
+    switch (lhs, rhs) {
+    case (.none, _):
+        return true
+    case (_, .none):
+        return true
+    case let (.some(lhs), .some(rhs)):
+        return lhs == rhs
+    default:
+        fatalError("Impossible")
     }
```
gwynne, line 74: 1. The comment is wrong - it effectively says "true if both are nil or both the same, false if only one is nil or they're unequal", but the actual semantics are "true unless both are non-nil _and_ unequal".
2. This is a _very_ rare case where I think a custom operator would actually be appropriate:
    ```swift
    infix operator ?==: ComparisonPrecedence
    
    @_transparent
    public func ?== <T: Equatable>(lhs: T?, rhs: T?) -> Bool {
        (lhs ?? rhs) == (rhs ?? lhs)
    }
    ```
    (How's that for a confusing-looking way to provide the desired semantics? 😁)

Lambdas/GHHooks/GHMiddleware.swift
```diff
@@ -57,15 +86,32 @@ actor GHMiddleware: ClientMiddleware {
...
             ])
 
-            return response
+            /// If this is not _the_ retry,
+            /// and if the authorization is set to use an authenticator,
+            /// and if the response status is `401 Unauthorized`,
+            /// then retry the request with a force-refreshed token.
+            if !isRetry,
+               case .authenticator = authorization,
+               response.statusCode == 401 {
```
gwynne, line 101: When possible, we should yank in the new `HTTPTypes` package so you can write `.unauthorized` instead of `401` here...

gwynne (approved):
A few nits, nothing I'm gonna force you to change (not even the actor locking logic, I've screwed up on that one too many times in a row as it is 😅)

Lambdas/GHHooks/Extensions/Embed+Equtable.swift
MahdiBM: Hmm i'd rather keep the switch just so it doesn't result in unnecessary equitability checks.

Lambdas/GHHooks/GHMiddleware.swift
MahdiBM: There are so many packages that can use the HTTPTypes package ... this is all from the OpenAPI-Generator so not exactly under my control but it's possible i do a PR for it someday.

Lambdas/GHHooks/ActorLock.swift
```diff
@@ -0,0 +1,22 @@
+
+actor ActorLock {
+    private var isLocked = false
+    private var lockWaiters: [CheckedContinuation<Void, Never>] = []
+
+    /// Acquires a lock and performs the async work, then releases the lock.
```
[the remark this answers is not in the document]
MahdiBM: > retried after failure
I think you created this comment in the wrong file? you mean in `GHMiddleware` i assume?
i guess i could add a comment

PR #70, opened 2023-07-18
Work around apple/swift#67214 (major compilation slowdown on Linux)

gwynne:
See https://github.com/apple/swift/issues/67214#issuecomment-1640129927 in particular.


MahdiBM (approved):
NICE!

PR #77, opened 2023-07-23
Rework Database

### LINKED ISSUE #75: GitHub linking should work even if the user doesn't exist yet
GitHub linking currently requires the user to exist, so they can link their account.
This assumes someone has already gotten a coin on Discord, so they are a known user.
It should instead just create a new user with 0 coins and successfully link their GitHub account.

MahdiBM:
Cleaning up a bunch of things and reworking the database.

the new users table will have better column names, and the coin entries are also stored on another database so they don't slowdown getting raw user objects.

also resolves #75.


Sources/Penny/Services/CoinService/DefaultCoinService.swift
```diff
@@ -69,64 +77,46 @@ actor DefaultCoinService: CoinService {
...
+        let url = "https://api.github.com/user/\(encodedID)"
+        logger.debug("Will make a request to get GitHub user name", metadata: [
+            "user": "\(user)",
+            "url": .string(url),
+        ])
+        var request = HTTPClientRequest(url: url)
+        request.headers = [
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": "2022-11-28",
+            "User-Agent": "Penny/1.0.0 (https://github.com/vapor/penny-bot)"
```
gwynne, line 104: I do not at all love that this is hardcoded. (Nor for that matter the GitHub API version.)

Sources/SharedServices/DynamoUserRepository.swift
```diff
@@ -14,113 +14,67 @@ struct DynamoUserRepository {
...
-    
-    init(db: DynamoDB, eventLoop: any EventLoop, logger: Logger) {
+    let tableName = "penny-user-table"
+    let discordIndex = "D-ID-GSI"
+    let githubIndex = "GH-ID-GSI"
+
+    init(db: DynamoDB, logger: Logger) {
         self.db = db
-        self.eventLoop = eventLoop
+        self.eventLoop = db.eventLoopGroup.any()
```
gwynne, line 28: Bonus 😍😍😍 for knowing to use `.any()` instead of `.next()`!!! ❤️

gwynne (approved):
As Fabian loves to say, giphy

One small nit, but nbd.

Sources/Penny/Services/CoinService/DefaultCoinService.swift
MahdiBM: What to do then?
It's not like it really matters to us.

MahdiBM:
The test failures are due to migration stuff. Before migration stuff CI was all green.

0xTim:
I wish I'd had a chance to review this PR. Unfortunately this breaks many of the concepts of single table design we were taking advantage of before. It's fine now but it's going to make it more complex to evolve and scale in the future

gwynne:
@0xTim Can you elaborate?

MahdiBM:
I see Googling "DynamoDB singe-table design" comes up with some results.
I'll give them a read sometime soon to see what's the idea behind it, and if the idea seems as good/interesting to me as it apparently is to Tim.

PR #80, opened 2023-07-24
Rework/Cleanup the main Discord-Penny target

MahdiBM:
More cleanup. Would specially be nice for better testability when it's done.


MahdiBM:
<not actually ready for review>

MahdiBM:
@gwynne Does this look _slightly_ better than before?
I removed all the "shared" instances.
This is just 1/3 of the work there .... I'll need to do another round of more cleanup and one last round to simplify the tests probably.

MahdiBM:
@gwynne need https://github.com/gwynne/swift-semver/pull/1 for full `-strict-concurrency=complete` compatibility.
Then i can update the PR.

Sources/Penny/+Array.swift
```diff
@@ -28,6 +28,12 @@ extension Array {
 }
 
 extension Array<String> {
+    mutating func appendUnique(_ element: String) {
```
gwynne, line 31: This is just _screaming_ out that what you really want here is `swift-collections`'s `OrderedSet`

gwynne (approved):
I'm hitting approve here despite the one nit... I really would be happier if you addressed it, but I'm leaving it up to you.

Sources/Penny/+Array.swift
MahdiBM: Hmmm good idea

PR #81, opened 2023-07-25
Replace Docs repo CI that files issues for translations

MahdiBM:


MahdiBM:
@gwynne review please 🙂

Lambdas/GHHooks/EventHandler/Handlers/DocsIssuer.swift
```diff
@@ -0,0 +1,135 @@
...
+                    metadata: ["number": .stringConvertible(pr.number)]
+                )
+                continue
+            }
+            try await fileIssue(number: pr.number)
+        }
+    }
+
+    /// Should not contain any labels that indicate no need for a new issue.
+    func isNotExemptFromNewIssue(pr: SimplePullRequest) -> Bool {
```
gwynne, line 55: This method name requires parsing a double-negative, which is both not idiomatic Swift and is just generally awkward. It's better to have to write `guard !isExemptFromNewIssue(...)` above than to do the extra mental gymnastics to make sense of the triple negative the guard introduces as things stand.

Lambdas/GHHooks/EventHandler/Handlers/DocsIssuer.swift
```diff
@@ -0,0 +1,135 @@
...
+                continue
+            }
+            let files = try await getPRFiles(number: pr.number)
+            /// PR must contain file changes for files that are in the `docs` directory.
+            /// Otherwise there is nothing to be translated and there is no need for a new issue.
+            guard files.contains(where: { file in
+                file.filename.hasPrefix("docs/") &&
+                [.added, .modified].contains(file.status)
+            }) else {
+                logger.warning(
```
gwynne, line 44: Is this really important enough to be a warning?

Lambdas/GHHooks/EventHandler/Handlers/DocsIssuer.swift
```diff
@@ -0,0 +1,135 @@
...
+        self.repo = try context.event.repository.requireValue()
+    }
+
+    func handle() async throws {
+        guard repo.id == Configuration.docsRepoID else {
+            return
+        }
+        for pr in try await getPRsRelatedToCommit() {
+            guard isNotExemptFromNewIssue(pr: pr) else {
+                logger.warning(
```
gwynne, line 31: This really feels like it should be a `.notice` at most.

Lambdas/GHHooks/EventHandler/Handlers/DocsIssuer.swift
```diff
@@ -0,0 +1,135 @@
...
+        - [ ] German
+        - [ ] Dutch
+        - [ ] Italian
+        - [ ] Spanish
+        - [ ] Polish
+        - [ ] Korean
+
+        Assigned to @vapor/translators - please submit a PR with the relevant updates and check the box once merged. Please ensure you tag your PR with the `translation-update` so it doesn't create a new issue!
+        """
+    }
```
gwynne, line 134: Yeah, no, I gotta put my foot down on this one. Even back when I was just tinkering with the existing workflow Tim pointed out that this needs to be a template (and in the docs repo) for the sake of others who shouldn't have to be able to parse Penny's code just to update the translation list (it happens more than you'd think).

gwynne (changes requested):
A few nits, one major concern.

Lambdas/GHHooks/EventHandler/Handlers/DocsIssuer.swift
MahdiBM: for debugging purposes ... at which point i guess i can put it at debug

Lambdas/GHHooks/EventHandler/Handlers/DocsIssuer.swift
MahdiBM: yeah i guess i'll change to debug

MahdiBM:
PR is ready but still need to change the branch to main branch when the docs PR is merged.

MahdiBM:
The `testLeafRenders()` function will fail until vapor/docs#867 is merged.

Lambdas/GHHooks/RenderClient/LeafEncoder.swift
0xTim, line 1: I wonder why this isn't in LeafKit - any idea @gwynne ?

Package.swift
```diff
@@ -224,8 +227,7 @@ let package = Package(
                 .product(name: "OpenAPIRuntime", package: "swift-openapi-runtime"),
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Atomics", package: "swift-atomics"),
-                .product(name: "Crypto", package: "swift-crypto"),
-                .product(name: "LeafKit", package: "leaf-kit")
+                .product(name: "Crypto", package: "swift-crypto")
```
0xTim, line 230:
```suggestion
                .product(name: "Crypto", package: "swift-crypto"),
```
Minor not - I prefer trailing commas in arrays to avoid this coming up as a change

0xTim (commented):
Overall LGTM - one nit and one query

Lambdas/GHHooks/RenderClient/LeafEncoder.swift
MahdiBM: should be honestly

Package.swift
MahdiBM: To be clear, haven't been doing it because if i wanted to do it i would have also liked to do it in a complete way and not half-way there.
Also swift-format has been on my to-do list which will solve the problem.

0xTim: Yep very happy to, ideally we'll roll it out across the org

Lambdas/GHHooks/RenderClient/LeafEncoder.swift
MahdiBM: The whole thing is internal as well, so shouldn't be a problem.

Package.swift
MahdiBM: sure. I'll try to see what Apple folks use. Then try on this repo, and eventually roll out everywhere.
When we want to roll out, we'll need to set up the config file in the repo, and @gwynne can also update the reusable CI.

gwynne (approved):
Looks fine 😛

PR #83, opened 2023-07-31
Implement /auto-faqs command

### LINKED ISSUE #19: Automatic messages for certain texts
Penny should be able to automatically respond with certain answers when it detects a certain sequence of words.
For example, we can set penny so if it sees `No custom working directory set for this scheme` in a message, it responds with `Take a look at this page to see how to set a custom working directory: [url]`.

### Logic
Apparently Slack has this feature built in it, you can also take a look at slack's approach if you want.
What I'm thinking about:
1- Penny receives a message and passed it to some kind of handler to take of the rest of the story.
2- The handler looks for any of those texts, and responds with the predefined answer if needed.

### Interface
We should use Slash commands to let certain people manage these commands. We can also set command's permissions so its only accessible by certain roles.
Those authorized users should be able to add, edit and remove. There should also be a list of these predefined texts.

MahdiBM:
Resolves #19.

- [x] Tests
- [ ] Update perms in AWS console


Sources/Penny/+String.swift
```diff
@@ -34,14 +34,21 @@ extension String {
...
             .lowercased()
             .folding(options: .diacriticInsensitive, locale: nil)
     }
-    
+
+    /// No whitespaces or lines and makes the string case, diacritic and punctuation insensitive.
+    func superHeavyFolded() -> String {
+        self.removingOccurrences(of: .whitespacesAndNewLinesAndPunctuations)
+            .lowercased()
+            .folding(options: .diacriticInsensitive, locale: nil)
```
gwynne, line 49: Ew. Don't use `CharacterSet` if you don't have to.
```suggestion
        self.filter { !($0.isWhitespace || $0.isNewline || $0.isPunctuation }
            .lowercased()
            .folding(options: .diacriticInsensitive, locale: nil)
```

Sources/Penny/Handlers/InteractionHandler.swift
```diff
@@ -451,6 +452,171 @@ private extension InteractionHandler {
...
+                    .requireComponent(customId: "expression")
+                    .requireTextInput()
+                    .value.requireValue()
+                let newValue = try modal.components
+                    .requireComponent(customId: "value")
+                    .requireTextInput()
+                    .value.requireValue()
+
+                if expression.contains("\n") {
+                    let expNoNewLines = expression.replacingOccurrences(of: "\n", with: " ")
```
gwynne, line 471:
```suggestion
                    let expNoNewLines = expression.replacing("\n", with: " ")
```

Sources/Penny/Handlers/InteractionHandler.swift
```diff
@@ -850,7 +1123,36 @@ private extension InteractionHandler {
...
+        let all = try await context.services.autoFaqsService.getAll().map(\.key)
+        let queried: ArraySlice<String>
+        if foldedExpression.isEmpty {
+            queried = ArraySlice(all.sorted { $0 > $1 }.prefix(25))
+        } else {
+            queried = all
+                .filter { $0.heavyFolded().contains(foldedExpression) }
+                .sorted { $0 > $1 }
+                .prefix(25)
+        }
```
gwynne, line 1145:
```suggestion
        let queried =
                (!foldedExpression.isEmpty ? all.heavyFolded().filter { $0.contains(foldedExpression) } : all)
                .sorted { $0 > $1 }
                .prefix(25)
```

gwynne (approved):
My usual bunch of minor nits, nothing worth being a blocker, up to you which suggestions you ignore.

Sources/Penny/+String.swift
MahdiBM: why though? :eyes:

Sources/Penny/Handlers/InteractionHandler.swift
MahdiBM: hmm if that's one of the new regex functions, it'll probably cause problems with "static-stdlib" flag 🤔

Sources/Penny/Handlers/InteractionHandler.swift
MahdiBM: I don't like the original code, but i don't think i like this new code either 😅

gwynne:
@MahdiBM Wanna toss in a `swift package update` while we're here? There's updates for `swift-cmark`, `swift-openapi-generator`, and `swift-openapi-runtime`.

MahdiBM:
Actually not a bad idea to not do that 😅 just so we test the new Swift dependabot stuff

MahdiBM:
Not sure if main isn't already updated

Sources/Penny/Handlers/InteractionHandler.swift
MahdiBM: I think this is a Set, not a String, so heavy folded is called for each string not each character 😅 the array slice probably threw you off

Sources/Penny/Handlers/InteractionHandler.swift
MahdiBM: I see there are regex variants too ... I'm not sure if it won't just infer to use the regex variant.
I'm going to give it a try.

PR #93, opened 2023-08-11
Update to Soto 7

MahdiBM:
Waiting on https://github.com/soto-project/soto-core/pull/581

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -89,7 +89,7 @@ let package = Package(
         .package(url: "https://github.com/DiscordBM/DiscordLogger.git", from: "1.0.0-rc.1"),
         /// Not-released area:
         .package(url: "https://github.com/soto-project/soto.git", from: "7.0.0-alpha.1"),
-        .package(url: "https://github.com/soto-project/soto-core.git", from: "7.0.0-alpha.1"),
+        .package(url: "https://github.com/soto-project/soto-core.git", from: "7.0.0-alpha.2"),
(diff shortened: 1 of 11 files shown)


Package.swift
```diff
@@ -153,7 +155,7 @@ let package = Package(
         .lambdaTarget(
             name: "AutoPings",
             additionalDependencies: [
-                .product(name: "SotoS3", package: "soto"),
+                .product(name: "SotoS3", package: "soto")
```
gwynne, line 158: Why? Trailing commas in array and dictionary literals are considered good form.

MahdiBM: Not sure how but swift-format somehow ran on a bunch of files and automatically formatted them for me

gwynne: For the most part I actually find the changes to be a marked improvement, but in this particular case I'd say a rule needs adjusting

MahdiBM: Wait now that i'm looking at it from PC...
This is actually completely expected formatting behavior. If there is only 1 element, it excludes the trailing comma.
Otherwise all elements including the last element will have a trailing comma in a collection.

gwynne: Still bad form IMO

MahdiBM: I can't see a settings for disabling that, but I don't mind it anyway.

MahdiBM:
(I know it won't auto-merge before someone approves)

gwynne (approved):

PR #96, opened 2023-08-13
Handle project-board tasks

### LINKED ISSUE #66: Replace the "projectboard" workflow currently present in all repos
Currently, every single Vapor repo contains a copy of one of the two versions of the `projectboard.yml` workflow (either [the original overcomplicated version](https://github.com/vapor/apns/blob/main/.github/workflows/projectboard.yml) or [the much-simplified one](https://github.com/vapor/redis/blob/main/.github/workflows/projectboard.yml) I wrote back in May). Both versions are attempts to deal with the same concern (with varying degrees of efficiency and correctness): the fact that [the built-in Projects automation only works for one repo at a time](https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/adding-items-automatically). Naturally, as a GitHub Actions workflow it still has to be applied on a repo-by-repo basis; one can leverage a reusable workflow to minimize duplication, but each repo still needs a copy of the `projectboard.yml`. The most significant problem with this is the one that's already obvious just from the fact that both versions are still in use: Every single one still has to be updated if the reusable workflow invocation has to change.

All of this is a complete waste of effort and CI minutes and actions queue slots. The same events that trigger the workflow are also already being sent to Penny's GitHub webhook lambda - to say nothing of the fact that Penny can implement complex logic based on those events far more effectively than the tortured excuse for decision-making that's all workflows are capable of without referring to external scripts (which come with their own problems).

Therefore, Penny should implement the semantics currently handled by [the reusable workflow that backs the per-repo files](https://github.com/vapor/ci/blob/reusable-workflows/.github/workflows/update-project-boards-for-issue.yml) (the workflow behind the old version takes the same actions, though it goes through a _lot_ more rigamarole to get there). Once that logic is deployed, the reusable workflow can be trivially disabled and the individual `projectboard.yml` files can be removed from each repo's `main` branch en masse (probably by script).

The (semi-)formal semantics currently provided by the reusable workflow and the third-party action it invokes are as follows:

1. On `issue.unlabeled` where `event.label.name` is one of `"help wanted"` or `"good first issue"`:
    a. Set `boards` to `["Help Wanted Issues"]` or `["Beginner Issues"]` respectively.
    b. For each `board` in `boards`, remove issue from `board` if present.
    c. Stop.
2. On `issue.labeled` where `event.label.name` is one of `"help wanted"` or `"good first issue"`:
    a. Set `boards` to `["Help Wanted Issues"]` or `["Beginner Issues"]` respectively. Go to step 4.
3. On `issue.assigned`, `.unassigned`, `.reopened`, `.closed`:
    a. Set `boards` to `[]`.
    b. If `event.issue.labels.contains { $0.name == "help wanted" }`, add "Help Wanted Issues" to `boards`.
    c. If `event.issue.labels.contains { $0.name == "good first issue" }`, add "Beginner Issues" to `boards`.
    d. Go to step 4.
4. If `event.issue.state == "closed"`, set `column` to "Done". Go to step 7.
5. If `!event.issue.assignees.compacted().isEmpty`, set `column` to "In Progress". Go to Step 7.
6. Set `column` to "To Do". Go to step 7.
7. Repeat for each `board` in `boards`:
    a. If issue present in `board`, move issue to `column`. Otherwise add issue to `column` in `board`.

Or in much simpler terms:

- When an issue has or gets one of the two labels, it should be added to the board for that label, if it isn't already there. If it later loses that label it should be removed from the corresponding board. An issue can have both labels and be on both boards.
- When an issue's added to a board, it should go in the appropriate column for its status and assignee(s).
- When an issue's status and/or assignee(s) change, it should be moved to the appropriate column of each board it's on as needed.
- An issue goes in the "Done" column if it's closed (regardless of assignee(s)), in the "In Progress" column if it's open and has at least one assignee, or in the "To Do" column for all other cases.

I should probably also mention I plan to migrate the existing boards (which are all "Classic" projects) to "modern" projects since IIRC classic and non-classic projects use different API endpoints.

### LINKED ISSUE #95: When editing or posting messages due to an issue being closed, Penny does not report who closed it
For example, Penny posted this message:
Screenshot 2023-08-13 at 04 45 29

Unless one stops to think about it, this reads a lot like it was BdeBock who actually closed the issue, while in fact they are the OP and it was @MahdiBM who closed it.

MahdiBM:
Resolves #66. 
Also i sneaked a "resolves #95" into the PR.


Lambdas/GHHooks/EventHandler/Handlers/IssueHandler.swift
gwynne, line 1: I can't comment in the file itself since it's not modified yet in this PR, but in [+TaskGroup.swift](https://github.com/vapor/penny-bot/tree/mmbm-project-board/Lambdas/GHHooks/Extensions/+TaskGroup.swift), may I suggest:
```swift
func withThrowingAccumulatingVoidTaskGroup(
    tasks: [@Sendable @escaping () async throws -> Void]
) async throws {
    /// TODO: change to discarding task-group when updated to Swift 5.9.
    /// N.B.: Despite the "Swift 5.9" notation on SE-0381, discarding task groups were actually
    /// available since 5.8.0, per https://github.com/apple/swift/pull/62963. This particular feature
    /// flag is the check the stdlib itself uses to determine whether to provide the API.
    #if $BuiltinCreateTaskGroupWithFlags
    try await withThrowingDiscardingTaskGroup { group in
        for task in tasks { group.addTask { try await task() } }
    }
    #else
    try await withThrowingTaskGroup(of: Void.self) { group in
        for task in tasks { group.addTask { try await task() } }
        var errors: [any Error] = []
        while let result = await group.nextResult() {
            switch result {
            case let .failure(error): errors.append(error)
            case .success: break
            }
        }
        guard errors.isEmpty else { throw Errors.multipleErrors(errors) }
    }
    #endif
}
```

MahdiBM: 👍, but I still need to catch the errors.
TaskGroups do wait for all tasks to finish in any case, but they don't throw the error unless you use one of the functions in the closure, like .waitForAll() IIRC.

gwynne: Not so for discarding task groups. Per stdlib docs (though I roll my eyes at the obvious lost train of thought near the end...):
```swift
/// Error Handling and Implicit Cancellation
/// ========================================
///
/// Since it is not possible to explicitly await individual task completions,
/// it is also not possible to "re-throw" an error thrown by one of the child
/// tasks using the same pattern as one would in a ``ThrowingTaskGroup``:
///
/// ```
/// // ThrowingTaskGroup, pattern not applicable to ThrowingDiscardingTaskGroup
/// try await withThrowingTaskGroup { group in
///   group.addTask { try boom() }
///   try await group.next() // re-throws "boom"
/// }
/// ```
///
/// Since discarding task groups don't have access to `next()`, this pattern
/// cannot be used.
/// Instead,
/// a *throwing discarding task group implicitly cancels itself whenever any
/// of its child tasks throws*.
///
/// The *first error* thrown inside such task group
/// is then retained and thrown
/// out of the `withThrowingDiscardingTaskGroup` method when it returns.
///
/// ```
/// try await withThrowingDiscardingTaskGroup { group in
///   group.addTask { try boom(1) }
///   group.addTask { try boom(2, after: .seconds(5)) }
///   group.addTask { try boom(3, after: .seconds(5)) }
/// }
/// ```
///
///
///
/// Generally, this suits the typical use-cases of a
/// discarding task group well, however, if you wanted to prevent specific
/// errors from cancelling the group
///
///
///
///
/// Throwing an error in one of the child tasks of a task group
/// doesn't immediately cancel the other tasks in that group.
/// However,
/// throwing out of the `body` of the `withThrowingTaskGroup` method does cancel
/// the group, and all of its child tasks.
```

MahdiBM: Ok then based on this comments we can't even use discarding task groups unlike my comments.

The whole point of what I do is to avoid losing any errors as well as each task not cancelling any other task if it throws.

gwynne: Why do you need the concurrent dispatch of tasks anyway? I feel like it'd be less error-prone to just do them sequentially, and it's not as if they're so time-consuming that- Ohhh, I get it, it's the "the concurrency must be structured" thing 🤣

MahdiBM: I didn't mention concurrency as any of the main points of using this 😅, but it should be better than not having the concurrency.

MahdiBM:
Shouldn't delete the branch after the PR merge because it's mentioned in https://github.com/apple/swift-package-manager/issues/6804.

gwynne:
> Shouldn't delete the branch after the PR merge because it's mentioned in [apple/swift-package-manager#6804](https://github.com/apple/swift-package-manager/issues/6804).

Inaccurate - you referred to the individual commit. I recommend you tag that commit (`git pull --tags && git tag -s -m 'Tag for apple/swift-package-manager#6804' spm-issue-6804 735869f555a305b8bbbad8ff4f7416c18bfae89e && git push --tags` (if you don't have GPG signing configured, use `-a` instead of `-s`) to make sure it doesn't get garbage-collected; it will then no longer be dependent on the branch to be kept "alive".

gwynne:
> ... it will then no longer be dependent on the branch to be kept "alive".

Oh, I should also mention that you can then change the URL in the issue- nevermind, you already did ❤️

Lambdas/GHHooks/Extensions/+PR.Label.swift
```diff
@@ -1,38 +1,38 @@
...
 
 extension PullRequest {
-    var knownLabels: [KnownLabel] {
-        self.labels.compactMap {
-            KnownLabel(rawValue: $0.name)
+
+    enum KnownLabel: String {
+        case semVerMajor = "semver-major"
+        case semVerMinor = "semver-minor"
+        case semVerPatch = "semver-patch"
```
gwynne, line 8: Fabian added a label `semver-noop` on PostgresNIO to mean the same thing we use `no-release-needed` for; in case we ever add any behavior for it (such as release notes in the style Fabian uses), we should probably recognize it as an alias.

Lambdas/GHHooks/Extensions/+Issue.Label.swift
```diff
@@ -0,0 +1,29 @@
+import GitHubAPI
+
+extension Issue {
+    enum KnownLabel: String {
+        case enhancement
+        case bug
```
gwynne, line 6: Might as well recognize `duplicate`, `wontfix`, and `invalid` here while we're at it.

Lambdas/GHHooks/EventHandler/Handlers/ProjectBoardHandler.swift
```diff
@@ -0,0 +1,249 @@
...
+        var alreadyMoved = false
+        for column in otherColumns {
+            let cards = try await cards(column: column)
+            if let card = cards.firstCard(note: note) {
+                if alreadyMoved {
+                    try await self.delete(cardID: card.id)
+                } else {
+                    try await move(cardID: card.id)
+                    alreadyMoved = true
+                }
```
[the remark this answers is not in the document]
MahdiBM: If the card exists in columns other than the target column,
If the card is alreadyMoved to the target column, delete the card from the current column,
If the card has not already moved to the target column, move it there.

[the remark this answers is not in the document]
gwynne: Wouldn't it be easier to just "find column card is currently in and move to target if needed"?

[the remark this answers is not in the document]
MahdiBM: Would certainly be easier. I just wanted to be safe.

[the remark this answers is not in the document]
gwynne: In this case, I think more understandable code is worth more than pedantic handling of the project boards (yes, _I'm_ saying to be less pedantic 😆); I'm not especially worried about the boards getting cluttered by duplicate cards.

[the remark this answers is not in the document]
MahdiBM: I don't have a big preference but rather keep it the way it is 😅

Lambdas/GHHooks/Extensions/Embed+Equtable.swift
```diff
@@ -63,6 +63,15 @@ infix operator ?==: ComparisonPrecedence
...
+private func ?== <E: Equatable>(lhs: E?, rhs: E?) -> Bool {
+    switch (lhs, rhs) {
+    case (.none, _):
+        return true
+    case (_, .none):
+        return true
+    case let (.some(lhs), .some(rhs)):
+        return lhs == rhs
+    default:
+        fatalError("Impossible")
```
gwynne, line 75:
```suggestion
```
Unnecessary; the compiler can correctly determine that the switch is already exhaustive.

gwynne (approved):
One minor nit, please do fix it, but approved anyway 🙂

Lambdas/GHHooks/Extensions/Embed+Equtable.swift
MahdiBM: Hmm I think i did expect it to but it couldn't

PR #101, opened 2023-08-18
Correctly check closed-by user for closed issues

MahdiBM:


gwynne (approved):
Fine, if you must. I think it's bizarre to deliberately have the bot making _more_ API requests instead of leveraging any chance to make fewer, but you do you... 😂

PR #106, opened 2023-08-28
Prefix Markdown Docs Based On Visible Text

MahdiBM:


Lambdas/GHHooks/Extensions/String+Document.swift
```diff
@@ -71,6 +80,21 @@ extension String {
         return lines.joined(separator: "\n")
     }
 
+    /// Doesn't count markdown attributes towards the limit.
+    @_disfavoredOverload
```
gwynne, line 84: Why is this attribute needed? (Disfavored overloads are notoriously finicky at best.)

Sources/Shared/+String.swift
```diff
@@ -6,12 +6,17 @@ extension String {
         ) ?? self
     }
 
-    public func unicodesPrefix(_ maxUnicodeScalars: Int) -> String {
+    @_disfavoredOverload
```
gwynne, line 9: Again, why?

gwynne (commented):
Overall, looks excellent! Just the two questions

Lambdas/GHHooks/Extensions/String+Document.swift
MahdiBM: The function names are the same.
Unless you can suggest a good function name for the functions that return both the remaining and the result, then we'll need to roll with this.

MahdiBM: I've basically duplicated the `unicodesPrefix()` function, one just returns the result, the other returns the remaining too.

gwynne (approved):

PR #119, opened 2023-09-30
Report StackOverflow questions

MahdiBM:

### DIFF
--- a/Sources/Penny/SOChecker.swift
+++ b/Sources/Penny/SOChecker.swift
@@ -22,13 +22,13 @@ actor SOChecker {
 
     nonisolated func run() {
         Task { [self] in
+            if Task.isCancelled { return }
             do {
                 try await self.check()
-                try await Task.sleep(for: .seconds(60 * 5)) /// 5 mins
             } catch {
                 logger.report("Couldn't check SO questions", error: error)
-                try await Task.sleep(for: .seconds(60 * 5))
             }
+            try await Task.sleep(for: .seconds(60 * 5)) /// 5 mins
             self.run()
         }
     }
--- a/Sources/Penny/SOChecker.swift
+++ b/Sources/Penny/SOChecker.swift
@@ -85,6 +85,17 @@ struct SOQuestions: Codable {
             let profileImage: String?
             let displayName: String
             let link: String?
+
+            enum CodingKeys: String, CodingKey {
+                case accountID = "account_id"
+                case reputation
+                case userID = "user_id"
+                case userType = "user_type"
+                case acceptRate = "accept_rate"
+                case profileImage = "profile_image"
+                case displayName = "display_name"
+                case link
+            }
         }
 
         let tags: [String]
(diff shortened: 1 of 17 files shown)


Sources/Penny/SOChecker.swift
```diff
@@ -0,0 +1,147 @@
...
+            enum CodingKeys: String, CodingKey {
+                case accountID = "account_id"
+                case reputation
+                case userID = "user_id"
+                case userType = "user_type"
+                case acceptRate = "accept_rate"
+                case profileImage = "profile_image"
+                case displayName = "display_name"
+                case link
+            }
```
gwynne, line 98: Suggestion - set `decoder.keyDecodingStrategy = .convertFromSnakeCase` on the `JSONDecoder` instead of manually specifying this.

Sources/Penny/SOChecker.swift
```diff
@@ -0,0 +1,147 @@
...
+
+actor SOChecker {
+
+    struct Storage: Sendable, Codable {
+        var lastCheckDate: Date? = nil
+    }
+
+    var storage = Storage()
+
+    let soService: any SOService
```
gwynne, line 14: It would be much preferable to use `some SOService` rather than `any SOService` if possible (not really relevant to performance in this case, but as a general rule `some` is always better than `any` if it's valid at all).

Sources/Penny/SOChecker.swift
```diff
@@ -0,0 +1,147 @@
+import Logging
+import DiscordBM
+import Markdown
+import Foundation
+
+actor SOChecker {
+
+    struct Storage: Sendable, Codable {
+        var lastCheckDate: Date? = nil
```
gwynne, line 9:
```suggestion
        var lastCheckDate: Date?
```
Explicit `nil` initialization of optionals is redundant.

Sources/Penny/SOChecker.swift
```diff
@@ -0,0 +1,147 @@
...
+        Task { [self] in
+            do {
+                try await self.check()
+                try await Task.sleep(for: .seconds(60 * 5)) /// 5 mins
+            } catch {
+                logger.report("Couldn't check SO questions", error: error)
+                try await Task.sleep(for: .seconds(60 * 5))
+            }
+            self.run()
+        }
```
gwynne, line 33: I get what this is doing and why, but it bugs me just the same, if only because you didn't leave yourself any outs by checking for cancellation. Also:
```suggestion
    nonisolated func run() {
        Task { [self] in
            do {
                try await self.check()
            } catch {
                logger.report("Couldn't check SO questions", error: error)
            }
            try await Task.sleep(for: .seconds(60 * 5)) /// 5 mins
            self.run()
        }
```
(And possibly put the catch part in `check()` instead, though I dunno if that really makes more sense or not.)

gwynne (changes requested):
Various minor stuff

Sources/Penny/SOChecker.swift
MahdiBM: eh, ok

Sources/Penny/SOChecker.swift
MahdiBM: Not possible because we can't mock it if we use `some`.

Sources/Penny/SOChecker.swift
MahdiBM: honestly i like this more for some reason 🤔 maybe i shouldn't

0xTim (approved):

gwynne (approved):

PR #129, opened 2023-10-15
CI improvements

MahdiBM:


0xTim (approved):

MahdiBM:
@gwynne do you approve now? 😛
I removed the wait-for-tests stuff.
Also @0xTim was not too happy with it either so i guess he should like it too.

.github/workflows/deploy-penny.yml
```diff
@@ -33,20 +38,23 @@ jobs:
           # be deployed to ECS.
           docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG -f Dockerfile .
           docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
-          echo "image=$ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG" >>"${GITHUB_OUTPUT}"
+          echo "image=$ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG" >> $GITHUB_OUTPUT
```
gwynne, line 41: Why?

gwynne (changes requested):

.github/workflows/deploy-penny.yml
MahdiBM: Probably my formatter? Not exactly sure

gwynne: Quoting the variable protects from spaces in the variable value. Adding the braces is syntactically purely stylistic in this context, but I prefer it just the same (it's "quoting" for variable names, so you could disambiguate, for example, `"$GITHUB_OUTPUT_FILE"` from `"${GITHUB_OUTPUT}_FILE"` - contrived but you get the idea).

MahdiBM: hmm i use it this way everywhere apparently so i guess i'll keep it this way.

PR #137, opened 2023-11-14
Fix Dependabot not updating indirect dependencies

MahdiBM:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,11 +1,34 @@
 version: 2
 enable-beta-ecosystems: true
 updates:
-  - package-ecosystem: "swift"
+  - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,11 +1,34 @@
...
       dependencies:
         patterns:
           - "*"
+  - package-ecosystem: "swift"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 6
+    allow:
+      - dependency-type: all
```
MahdiBM, line 18: This was the problem. I think they switched/fixed the behavior of it in "swift" ecosystem mid-day, which broke our stuff. It's still in beta so we can't complain too much, and it also seems like the default behavior of Dependabot for other ecosystems too, and they probably wanted it to be the same for Swift.

gwynne (approved):

PR #149, opened 2023-12-15
Minor refinements

MahdiBM:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,7 +45,7 @@ RUN find -L "$(swift build --package-path /build -c release --show-bin-path)/" -
 # ================================
 # Run image
 # ================================
-FROM ubuntu:latest
+FROM ubuntu:jammy
 
 # Make sure all system packages are up to date, and install only essential packages.
 RUN export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true \
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,11 +3,11 @@
 # ================================
 FROM swift:5.9-jammy as build
 
-# Install updates
+# Install OS updates
 RUN export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true \
     && apt-get -q update \
     && apt-get -q dist-upgrade -y \
-    && rm -rf /var/lib/apt/lists/*
+    && apt-get install -y libjemalloc-dev
 
 # Set up a build area
 WORKDIR /build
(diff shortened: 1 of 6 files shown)


0xTim (approved):

Dockerfile
```diff
@@ -17,37 +17,48 @@ WORKDIR /build
...
 RUN find -L "$(swift build --package-path /build -c release --show-bin-path)/" -regex '.*\.resources$' -exec cp -Ra {} ./ \;
 
-# Move /Templates/Penny directory to staging area
-RUN mkdir -p ./Templates && mv /build/Templates/Penny ./Templates/Penny && chmod -R a-w ./Templates/Penny;
-
 # ================================
 # Run image
 # ================================
-FROM ubuntu:jammy
+FROM ubuntu:latest
```
gwynne, line 48: This is not safe. Swift does not officially support post-Jammy Ubuntu yet.

Dockerfile
```diff
@@ -17,37 +17,48 @@ WORKDIR /build
...
 COPY . .
 
-# Build everything, with optimizations
-RUN swift build -c release --static-swift-stdlib --product Penny
+# Build everything, with optimizations, with static linking, and using jemalloc
+# N.B.: The static version of jemalloc is incompatible with the static Swift runtime.
+RUN swift build -c release \
+      --target Penny \
+      --static-swift-stdlib \
+      -Xlinker -ljemalloc
```
MahdiBM, line 31: `--target Penny` instead of `--product` because i removed the product as it seemed unnecessary.

Dockerfile
```diff
@@ -59,12 +70,14 @@ WORKDIR /app
...
+
 # Ensure all further commands run as the vapor user
 USER vapor:vapor
 
 # Let Docker bind to port 8080
 EXPOSE 8080
 
-# Start the Vapor service when the image is run, default to listening on 8080 in production environment
-ENTRYPOINT [ "./Penny" ]
-CMD [ "serve", "--env", "production", "--hostname", "0.0.0.0", "--port", "8080" ]
```
MahdiBM, line 70: `CMD` is useful for Vapor apps but doesn't do anything for Penny as it's not a Vapor app.

Lambdas/GHHooks/EventHandler/Handlers/IssueHandler.swift
```diff
@@ -228,12 +228,10 @@ private enum Status: String {
...
-        if issue.state_reason == .not_planned {
-            if issue.knownLabels.contains(.duplicate) {
-                self = .duplicate
-            } else {
-                self = .notPlanned
-            }
+        if issue.knownLabels.contains(.duplicate) {
+            self = .duplicate
+        } else if issue.state_reason == .not_planned {
+            self = .notPlanned
```
MahdiBM, line 234: if `duplicate` label exists then it's always a duplicate issue. Regardless if someone closed the issue as "Not Planned" or not.

Dockerfile
MahdiBM: Hmm i believe i only copied it from the new template Dockerfile... i was thinking the same too about Swift support 🤔

MahdiBM: reverted

gwynne (approved):

Lambdas/GitHubAPI/openapi-generator-config.yml
```diff
@@ -1,3 +1,4 @@
 generate:
   - types
   - client
+accessModifier: package
```
MahdiBM, line 4: They changed the default to `internal` before the 1.0.0 release.

MahdiBM:
Nightly failures are due to the upcoming features i enabled (internal import by default).
I'll take care of it when it's in a tagged-nightly and is more stable.

PR #155, opened 2024-01-06
Another try to fix duplicate ticket reports on Discord

MahdiBM:

### DIFF
--- a/Lambdas/GHHooks/EventHandler/Handlers/IssueHandler.swift
+++ b/Lambdas/GHHooks/EventHandler/Handlers/IssueHandler.swift
@@ -55,7 +55,9 @@ struct IssueHandler: Sendable {
     }
 
     func onEdited() async throws {
-        try await self.makeReporter().reportEdition()
+        try await self.makeReporter().reportEdition(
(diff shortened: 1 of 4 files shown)


Lambdas/GHHooks/EventHandler/Handlers/IssueHandler.swift
```diff
@@ -55,7 +55,9 @@ struct IssueHandler: Sendable {
     }
 
     func onEdited() async throws {
-        try await self.makeReporter().reportEdition()
+        try await self.makeReporter().reportEdition(
+            requiresPreexistingReport: self.action == .labeled
```
MahdiBM, line 59: Require a pre-existing-report for `labeled` events because otherwise they can cause a duplicate-discord-message problem because GitHub sends the `created` and `labeled` events too close to each other.

Technically all events other than `created` should require a preexisting message, but we don't require that because there are some tickets that were created before addition of this feature to Penny, so Penny never had the chance to create a Discord message for them when the ticket was `created`.

0xTim (approved):

Issue #100, opened 2023-08-18
Even better markdown handling


MahdiBM:
Penny makes a lot of effort to show GitHub markdown strings in Discord, in a form that is nicer to see in Discord.
But i still want to make it even better:

- [ ] Penny should convert strings like `#55` to #55 which is also a link, when sending to Discord.
- [ ] Also should count the actual string value of only `Text` elements when doing `String.unicodesPrefix()` for markdown documents.
  - [ ] This might need some PRs in `swift-markdown` to make it more flexible and nicer to use in Penny.

gwynne:
You're just lucky SwiftMarkdown is built directly atop cmark-gfm (which is literally GitHub's Markdown flavor). That being said, I don't think the automagic conversion of issue/PR numbers to links etc. thing is all that important; they're very context-dependent and don't always work even on GitHub in places you might think they should. Also no one on Discord generally cares.

MahdiBM:
I care though 🫣

I know not many people care, but if I were to work on penny with that assumption, I should've stopped a long time ago 😅

PR #165, opened 2024-01-27
Interpret issue/pr links like GitHub does in markdown

MahdiBM:
Partially resolves the first item of this issue: https://github.com/vapor/penny-bot/issues/100


Lambdas/GHHooks/Extensions/String+Document.swift
```diff
@@ -15,8 +15,17 @@ extension String {
         assert(hardLimit > 0, "Hard length limit must be greater than zero (got \(hardLimit)).")
         assert(hardLimit >= maxVisualLength, "maxVisualLength '\(maxVisualLength)' can't be more than hardLimit '\(hardLimit)'.")
 
+        /// Interpret urls like GitHub does.
+        /// For example GitHub changes `https://github.com/vapor/penny-bot/issues/99` to
+        /// `[vapor/penny-bot#99](https://github.com/vapor/penny-bot/issues/99)` which
+        /// ends up looking like a blue `vapor/penny-bot#99` text linked to the url.
+        let regex = #/https://github\.com/(?<org>[^/<>#%"{}|^\[\]\s]+)/(?<repo>[^/<>#%"{}|^\[\]\s]+)/(pull|issues)/(?<number>\d+)/#
```
gwynne, line 22: If you're gonna use the `#/ /#` "automatic extended syntax" syntax (no, that's not ungrammatical), at least take advantage of it to make your regex remotely readable. Also your two path components are too permissive; per GitHub's UI,

> The [organization] name may only contain alphanumeric characters or single hyphens, and cannot begin or end with a hyphen.

> The repository name can only contain ASCII letters, digits, and the characters ., -, and _.

Among other things. Basically, see below:

```suggestion
        let regex = #/
            https://(?:www\.)?github\.com
            /(?<org>[A-Za-z0-9](?:[A-Za-z0-9\-]*[A-Za-z0-9])?)
            /(?<repo>[A-Za-z0-9.\-_]+)
            /(?:pull|issues)
            /(?<number>\d+)
        /#
```

gwynne (changes requested):
You didn't really think I'd let you get away with it that easily, did you?

gwynne (approved):

PR #176, opened 2024-03-10
Use NIO singletons

MahdiBM:

### DIFF
--- a/Sources/Penny/Penny.swift
+++ b/Sources/Penny/Penny.swift
@@ -11,10 +11,13 @@ struct Penny {
     static func start(mainService: any MainService) async throws {
         /// Use `1` instead of `System.coreCount`.
         /// This is preferred for apps that primarily use structured concurrency.
-        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
         let httpClient = HTTPClient(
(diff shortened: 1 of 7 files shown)


gwynne (approved):

Sources/Penny/Penny.swift
```diff
@@ -11,10 +11,13 @@ struct Penny {
     static func start(mainService: any MainService) async throws {
         /// Use `1` instead of `System.coreCount`.
         /// This is preferred for apps that primarily use structured concurrency.
-        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
         let httpClient = HTTPClient(
-            eventLoopGroupProvider: .shared(eventLoopGroup),
-            configuration: .init(decompression: .enabled(limit: .size(1 << 32)))
+            eventLoopGroup: MultiThreadedEventLoopGroup.singleton,
```
MahdiBM, line 15: `HTTPClient`'s `.init()` does automatically use a singleton but it's definition is different from the ones used in NIO and sometimes choses to use the NIOTS ELG. That's why i've preferred to explicitly declare the ELG used by the `HTTPClient`.

MahdiBM: @gwynne FYI

PR #183, opened 2024-04-06
Minor message refinements

MahdiBM:

### DIFF
--- a/Sources/Penny/Handlers/MessageDeleteHandler.swift
+++ b/Sources/Penny/Handlers/MessageDeleteHandler.swift
@@ -69,7 +69,7 @@ private extension Payloads.CreateMessage {
                 timestamp: messageCreate.timestamp.date,
                 color: .red,
                 footer: .init(
-                    text: "From @\(member?.uiName ?? author.uiName)",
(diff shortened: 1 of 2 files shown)


Sources/Penny/Handlers/MessageDeleteHandler.swift
```diff
@@ -70,7 +69,7 @@ private extension Payloads.CreateMessage {
                 timestamp: messageCreate.timestamp.date,
                 color: .red,
                 footer: .init(
-                    text: "From @\(username)",
+                    text: "From @\(member?.uiName ?? author.uiName)",
```
MahdiBM, line 72: Mentions the nickname which can be the same as the username, but not always.

Sources/Penny/Handlers/MessageDeleteHandler.swift
```diff
@@ -80,8 +79,8 @@ private extension Payloads.CreateMessage {
                         inline: true
                     ),
                     .init(
-                        name: "Author ID",
-                        value: author.id.rawValue,
+                        name: "Author Username",
+                        value: author.username,
```
MahdiBM, line 83: Mentions the username

PR #186, opened 2024-04-12
Use `HTTPClient.shared` more

MahdiBM:

### DIFF
--- a/Lambdas/GHOAuth/OAuthLambda.swift
+++ b/Lambdas/GHOAuth/OAuthLambda.swift
@@ -19,7 +19,7 @@ struct GHOAuthHandler: LambdaHandler {
     typealias Event = APIGatewayV2Request
     typealias Output = APIGatewayV2Response
 
-    let client: HTTPClient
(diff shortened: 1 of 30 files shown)


Lambdas/GHOAuth/OAuthLambda.swift
```diff
@@ -38,10 +38,10 @@ struct GHOAuthHandler: LambdaHandler {
     }
 
     init(context: LambdaInitializationContext) async throws {
-        self.client = HTTPClient(eventLoopGroupProvider: .shared(context.eventLoop))
+        self.client = HTTPClient.shared
```
gwynne, line 41: Why do you even need a `client` property at all at this point?

gwynne (approved):

Lambdas/GHOAuth/OAuthLambda.swift
MahdiBM: I did think about it ... was kind of scared to remove all of them but I still can't see why Penny would need them ...

PR #189, opened 2024-04-21
Fix fetch-contributors logic to thank new contributors

MahdiBM:
Apparently it only fetched first 30?! not sure if GitHub recently changed this or not.

### DIFF
--- a/Lambdas/GHHooks/EventHandler/Handlers/ReleaseMaker.swift
+++ b/Lambdas/GHHooks/EventHandler/Handlers/ReleaseMaker.swift
@@ -331,13 +331,13 @@ struct ReleaseMaker {
 
         if case let .ok(ok) = response,
            case let .json(json) = ok.body {
-            logger.debug("Found some contributors", metadata: [
+            logger.debug("Fetched some contributors", metadata: [
                 "page": .stringConvertible(page),
                 "count": .stringConvertible(json.compactMap(\.id).count)
             ])
             return json
         } else {
-            logger.warning("Could not find more contributors", metadata: [
+            logger.warning("Error when fetching contributors", metadata: [
                 "page": .stringConvertible(page),
                 "response": "\(response)"
             ])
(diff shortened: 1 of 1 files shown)


Lambdas/GHHooks/EventHandler/Handlers/ReleaseMaker.swift
```diff
@@ -299,21 +297,51 @@ struct ReleaseMaker {
     }
 
     func getExistingContributorIDs() async throws -> Set<Int> {
+        var page = 1
+        var contributorIds: [Int] = []
+        /// Hack: Vapor has around this amount of contributors and we know it, so better
+        /// to reserve enough capacity for it up-front.
+        contributorIds.reserveCapacity(250)
+        while let contributors = try await self.getExistingContributorIDs(page: page) {
+            let ids = (consume contributors).compactMap(\.id)
```
MahdiBM, line 306: btw i don't think these `consume`s do much since the type of the variable is not `NonCopyable`. At least not based on the latest info i have. Maybe later they'll implement something to ensure e.g. COW is skipped when using these keywords.

Lambdas/GHHooks/EventHandler/Handlers/ReleaseMaker.swift
```diff
@@ -299,21 +297,61 @@ struct ReleaseMaker {
...
+        var contributorIds: [Int] = []
+        /// Hack: Vapor has around this amount of contributors and we know it, so better
+        /// to reserve enough capacity for it up-front.
+        contributorIds.reserveCapacity(250)
+        while true {
+            let (ids, hasNext) = try await self.getExistingContributorIDs(page: page)
+            if ids.isEmpty {
+                break
+            }
+            contributorIds.append(contentsOf: consume ids)
```
MahdiBM, line 310: btw i don't think these `consume`s do much since the type of the variable is not `NonCopyable`. At least not based on the latest info i have. Maybe later they'll implement something to ensure e.g. COW is skipped when using these keywords.

PR #193, opened 2024-05-19
Revert usage of `HTTPClient.shared`

MahdiBM:
Revert commits related to using `HTTPClient.shared`.
I don't think using the shared HTTPClient is necessary for Penny. It only moves Penny out of the control of the `HTTPClient`.

### DIFF
--- a/Lambdas/AutoFaqs/AutoFaqsHandler.swift
+++ b/Lambdas/AutoFaqs/AutoFaqsHandler.swift
@@ -1,8 +1,10 @@
 import AWSLambdaRuntime
 import AWSLambdaEvents
+import AsyncHTTPClient
 import Foundation
 import SotoCore
 import Models
+import Shared
 import LambdasShared
 
 @main
@@ -14,11 +16,12 @@ struct AutoFaqsHandler: LambdaHandler {
     let autoFaqsRepo: S3AutoFaqsRepository
 
     init(context: LambdaInitializationContext) async {
-        let awsClient = AWSClient(
-            httpClientProvider: .createNewWithEventLoopGroup(context.eventLoop)
+        let httpClient = HTTPClient(
+            eventLoopGroupProvider: .shared(context.eventLoop),
+            configuration: .forPenny
         )
-        self.awsClient = awsClient
-        self.autoFaqsRepo = S3AutoFaqsRepository(awsClient: awsClient, logger: context.logger)
+        self.awsClient = AWSClient(httpClient: httpClient)
+        self.autoFaqsRepo = S3AutoFaqsRepository(awsClient: self.awsClient, logger: context.logger)
     }
 
     func handle(
--- a/Lambdas/AutoFaqs/AutoFaqsHandler.swift
+++ b/Lambdas/AutoFaqs/AutoFaqsHandler.swift
@@ -14,7 +14,10 @@ struct AutoFaqsHandler: LambdaHandler {
     let autoFaqsRepo: S3AutoFaqsRepository
 
     init(context: LambdaInitializationContext) async {
-        self.awsClient = AWSClient()
+        let awsClient = AWSClient(
+            httpClientProvider: .createNewWithEventLoopGroup(context.eventLoop)
+        )
+        self.awsClient = awsClient
         self.autoFaqsRepo = S3AutoFaqsRepository(awsClient: awsClient, logger: context.logger)
     }
 
--- a/Sources/Penny/Penny.swift
+++ b/Sources/Penny/Penny.swift
@@ -9,12 +9,38 @@ struct Penny {
     }
 
     static func start(mainService: any MainService) async throws {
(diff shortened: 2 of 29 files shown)


Lambdas/AutoFaqs/AutoFaqsHandler.swift
```diff
@@ -14,8 +16,12 @@ struct AutoFaqsHandler: LambdaHandler {
     let autoFaqsRepo: S3AutoFaqsRepository
 
     init(context: LambdaInitializationContext) async {
-        self.awsClient = AWSClient()
-        self.autoFaqsRepo = S3AutoFaqsRepository(awsClient: awsClient, logger: context.logger)
+        let httpClient = HTTPClient(
```
0xTim, line 19: Do we need to worry about shutting this down?

MahdiBM: I don't think so considering it'll only need to shutdown when the process is killed.

MahdiBM: See: https://swift-open-source.slack.com/archives/C5GMX4LP9/p1682667872540799?thread_ts=1682620274.444989&cid=C5GMX4LP9

Sources/Penny/Penny.swift
```diff
@@ -9,12 +10,34 @@ struct Penny {
...
+            eventLoopGroup: MultiThreadedEventLoopGroup.singleton,
+            configuration: .forPenny
+        )
+        let awsClient = AWSClient(httpClient: httpClient)
 
-        let bot = try await mainService.makeBot()
+        /// These shutdown calls are only useful for tests where we call `Penny.main()` repeatedly
+        defer {
+            /// Shutdown in reverse order of dependance.
+            try! awsClient.syncShutdown()
```
0xTim, line 24: We should not be doing this - this is a Swift bug in that it's not showing you a warning but these methods are (or should be) marked with `noasync` - we need to call the async shutdowns, which unfortunately means no defers

0xTim (changes requested):

Sources/Penny/Penny.swift
MahdiBM: Removed all `syncShutdown()`, let's see how the tests go.

Sources/Penny/Penny.swift
```diff
@@ -23,5 +39,10 @@ struct Penny {
         for await event in await bot.events {
             EventHandler(event: event, context: context).handle()
         }
+
+        /// These shutdown calls are only useful for tests where we call `Penny.main()` repeatedly
+        /// Shutdown in reverse order of dependance.
+        try! await awsClient.shutdown()
+        try! await httpClient.shutdown()
```
0xTim, line 46: Any reason for force try these?

Sources/Penny/Services/AutoFaqsService/DefaultAutoFaqsService.swift
```diff
@@ -49,7 +49,7 @@ actor DefaultAutoFaqsService: AutoFaqsService {
         }
     }
 
-    let httpClient: HTTPClient = .shared
+    var httpClient: HTTPClient!
```
0xTim, line 52: Why is this an IUO

Sources/Penny/Services/FaqsService/DefaultFaqsService.swift
```diff
@@ -10,7 +10,7 @@ import NIOHTTP1
 
 actor DefaultFaqsService: FaqsService {
 
-    let httpClient: HTTPClient = .shared
+    var httpClient: HTTPClient!
```
0xTim, line 13: IUO?

Tests/PennyTests/Tests/LeafRenderTests.swift
```diff
@@ -10,24 +10,34 @@ import NIOPosix
...
+            on: httpClient.eventLoopGroup.next()
         )
     )
 
     override func setUp() async throws {
         FakeResponseStorage.shared = FakeResponseStorage()
     }
 
+    override func tearDown() async throws {
+        try! await httpClient.shutdown()
```
0xTim, line 38:
```suggestion
        try await httpClient.shutdown()
```

0xTim (approved):
A few comments but nothing blocking

Sources/Penny/Penny.swift
MahdiBM: not really, just copy-pasta.

Sources/Penny/Services/FaqsService/DefaultFaqsService.swift
MahdiBM: There is a reason although i don't remember it off-hand.

MahdiBM: They are likely initialized first, then populated with the http client.

PR #246, opened 2024-09-29
Move to SwiftTesting

MahdiBM:

### DIFF
--- a/Tests/PennyTests/Tests/+XCTest.swift
+++ b/Tests/PennyTests/Tests/+XCTest.swift
@@ -7,8 +7,7 @@ import Testing
 func expectMultilineStringsEqual(
     _ expression1: String,
     _ expression2: String,
-    file: StaticString = #filePath,
-    line: UInt = #line
+    sourceLocation: Testing.SourceLocation = #_sourceLocation
 ) {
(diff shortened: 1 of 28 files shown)


Tests/PennyTests/Tests/+XCTest.swift
0xTim, line 1: Rename this file as it's no longer XCTest based

Tests/PennyTests/Tests/LeafRenderTests.swift
```diff
@@ -30,20 +31,20 @@ class LeafRenderTests: XCTestCase {
...
-    override func setUp() async throws {
-        FakeResponseStorage.shared = FakeResponseStorage()
-    }
-
-    override func tearDown() async throws {
-        try await httpClient.shutdown()
+    init() {}
+    
+    deinit {
+        try! httpClient.syncShutdown()
```
0xTim, line 37: We should move to a `with` style pattern to manage the lifecycle of the HTTPClient - the deinit pattern is flaky and doesn't play nice with blocking or isolation

MahdiBM: Just used the shared `HTTPClient` in tests.

0xTim (approved):

PR #253, opened 2024-10-11
JWTKit 5, FoundationEssentials-only + more

MahdiBM:


Lambdas/GHOAuth/Models/GHOAuthPayload.swift
```diff
@@ -1,5 +1,6 @@
 import DiscordBM
 import JWTKit
+import struct Foundation.Date
```
0xTim, line 3: Shouldn't this be `FoundationEssentials.Date` to avoid linking all of Foundation?

Sources/Penny/Handlers/CoinFinder.swift
```diff
@@ -1,4 +1,4 @@
-import Foundation
+import Foundation /// Need to import the whole thing for the `String.split()` function call below?
```
0xTim, line 1: It should be in essentials unless it relies on ICU to ensure unicode is split correctly but I don't think that's the case, I thought that was all reimplemented in Swift for this reason

0xTim (approved):
Few comments to improve things but LGTM

Sources/Penny/Handlers/CoinFinder.swift
MahdiBM: Not sure what exactly is going on but things don't compile without importing the whole thing.
Something about `CharacterSet` I think, for one.

MahdiBM: Ah right IIRC `CharacterSet` is still not ported over? There were some issues around it in the main OpenAPIGenerator repo.

MahdiBM: This is my own comment from a few months ago. Apparently it was accurate:
https://github.com/apple/swift-openapi-runtime/pull/107#issuecomment-2181552319

MahdiBM: Yes, the problem was `CharacterSet`: https://github.com/vapor/penny-bot/pull/253/commits/e96f9918926531791ec2d870d65c5a1deedc02b5

PR #254, opened 2024-10-12
`SwiftReleasesChecker` to report new Swift releases

MahdiBM:


Sources/Penny/SwiftReleasesChecker.swift
```diff
@@ -0,0 +1,85 @@
...
+    let discordService: DiscordService
+    let logger = Logger(label: "SwiftReleasesChecker")
+
+    init(swiftReleasesService: any SwiftReleasesService, discordService: DiscordService) {
+        self.swiftReleasesService = swiftReleasesService
+        self.discordService = discordService
+    }
+
+    nonisolated func run() {
+        Task { [self] in
```
0xTim, line 27: Why are we doing this with unstructured concurrency?

MahdiBM: I was thinking about this. I should move things to ServiceLifecycle, but that'll have to be for another PR.

MahdiBM: Everything matches ServiceLifecycle too, coincidentally. Even the function name, other than the fact that the function is a "blocking" function which is what ServiceLifecycle expects.

MahdiBM: This pattern is what we use everywhere in Penny so can't just fix it in one place, Will need more changes.

MahdiBM: https://github.com/vapor/penny-bot/pull/255

0xTim (approved):

PR #255, opened 2024-10-13
Use Structured Concurrency

MahdiBM:
* Use `ServiceLifecycle`
* Remove all occurrences of unstructured `Task {}`s

### DIFF
--- a/Sources/Penny/BotStateManager.swift
+++ b/Sources/Penny/BotStateManager.swift
@@ -28,7 +28,7 @@ import Logging
 actor BotStateManager {
 
     let id = Int(Date().timeIntervalSince1970)
-    let services: HandlerContext.Services
+    let context: HandlerContext
     let disableDuration: Duration
     let logger: Logger
 
@@ -36,10 +36,10 @@ actor BotStateManager {
     var onStarted: (() async -> Void)?
 
     init(
-        services: HandlerContext.Services,
+        context: HandlerContext,
         disabledDuration: Duration = .seconds(3 * 60)
     ) {
-        self.services = services
+        self.context = context
         self.disableDuration = disabledDuration
         var logger = Logger(label: "BotStateManager")
         logger[metadataKey: "id"] = "\(self.id)"
@@ -87,7 +87,7 @@ actor BotStateManager {
 
     private func shutdown() {
         Task {
-            await services.cachesService.gatherCachedInfoAndSaveToRepository()
+            await context.cachesService.gatherCachedInfoAndSaveToRepository()
             await send(.didShutdown)
             self.canRespond = false
 
@@ -102,7 +102,7 @@ actor BotStateManager {
             if canRespond {
                 logger.warning("Received a did-shutdown signal but Cache is already populated")
             } else {
-                await services.cachesService.getCachedInfoFromRepositoryAndPopulateServices()
+                await context.cachesService.getCachedInfoFromRepositoryAndPopulateServices()
                 await startAllowingResponses()
             }
         }
@@ -115,7 +115,7 @@ actor BotStateManager {
 
     private func send(_ signal: StateManagerSignal) async {
         let content = makeSignalMessage(text: signal.rawValue, id: self.id)
-        await services.discordService.sendMessage(
+        await context.discordService.sendMessage(
             channelId: Constants.Channels.botLogs.id,
             payload: .init(content: content)
         )
--- a/Sources/Penny/BotStateManager.swift
+++ b/Sources/Penny/BotStateManager.swift
@@ -152,8 +152,8 @@ actor BotStateManager: Service {
 
     private func send(_ signal: StateManagerSignal) async {
         switch Constants.deploymentEnvironment {
-        case .local: break
-        case .testing, .prod: return
+        case .local: return
+        case .testing, .prod: break
         }
 
         let content = makeSignalMessage(text: signal.rawValue, id: self.id)
--- a/Sources/Penny/BotStateManager.swift
+++ b/Sources/Penny/BotStateManager.swift
@@ -50,7 +50,7 @@ actor BotStateManager: Service {
         switch Constants.deploymentEnvironment {
         case .local:
             break
-        case .prod:
+        case .testing, .prod:
             self.context.backgroundProcessor.process {
                 await self.cancelIfCachePopulationTakesTooLong()
             }
@@ -65,7 +65,9 @@ actor BotStateManager: Service {
     }
 
     func addCachesPopulationContinuation(_ cont: CheckedContinuation<Void, Never>) {
-        guard Constants.deploymentEnvironment == .prod else {
+        switch Constants.deploymentEnvironment {
+        case .local: break
+        case .testing, .prod:
             cont.resume()
             return
         }
@@ -88,7 +90,9 @@ actor BotStateManager: Service {
     }
 
     func canRespond(to event: Gateway.Event) async -> Bool {
-        if Constants.deploymentEnvironment == .prod {
+        switch Constants.deploymentEnvironment {
+        case .local: break
+        case .testing, .prod:
             await checkIfItsASignal(event: event)
         }
         return canRespond
@@ -147,7 +151,10 @@ actor BotStateManager: Service {
     }
 
     private func send(_ signal: StateManagerSignal) async {
-        guard Constants.deploymentEnvironment == .prod else { return }
+        switch Constants.deploymentEnvironment {
+        case .local: break
+        case .testing, .prod: return
+        }
 
         let content = makeSignalMessage(text: signal.rawValue, id: self.id)
         await context.discordService.sendMessage(
--- a/Sources/Penny/BotStateManager.swift
+++ b/Sources/Penny/BotStateManager.swift
@@ -48,26 +48,30 @@ actor BotStateManager {
 
     func start(onStarted: @Sendable @escaping () async -> Void) async {
         self.onStarted = onStarted
-        Task { await send(.shutdown) }
-        cancelIfCachePopulationTakesTooLong()
-    }
-
-    private func cancelIfCachePopulationTakesTooLong() {
-        Task {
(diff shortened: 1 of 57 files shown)


Sources/Penny/BotStateManager.swift
```diff
@@ -26,96 +26,121 @@ import Logging
...
+        case .local:
+            break
+        case .prod:
+            self.context.backgroundProcessor.process {
+                await self.cancelIfCachePopulationTakesTooLong()
+            }
+            self.context.backgroundProcessor.process {
+                await self.send(.shutdown)
+            }
+        }
```
MahdiBM, line 59: Some env checking to make sure i can manually run Penny without worrying about interrupting the prod penny.
Other than that the file changes just move stuff to structured concurrency.

Sources/Penny/Services/AutoFaqsService/DefaultAutoFaqsService.swift
```diff
@@ -165,32 +166,32 @@ actor DefaultAutoFaqsService: AutoFaqsService {
...
                 /// If canceled, set up the task again.
                 /// This way, the functions above can cancel this when they've got fresh items
                 /// and this will just reschedule itself for a later time.
-                self.setUpResetItemsTask()
+                await self.setUpResetItemsTask()
             }
         }
+        self.resetItemsTask = task
+        await task.value
     }
```
MahdiBM, line 195: this is task, but not unstructured.

Sources/Penny/WaiterService.swift
```diff
@@ -0,0 +1,47 @@
...
+    ///   - logger: A logger to log with.
+    ///   - backgroundProcessor: To process the continuation with.
+    ///   - passContinuation: Passes continuation using this closure to any other service you'd like.
+    ///   Then the other service is responsible for correctly and timely resolving the continuation.
+    init(
+        underlyingService: UnderlyingService,
+        logger: Logger = Logger(label: _typeName(Self.self)),
+        processingOn backgroundProcessor: BackgroundProcessor,
+        passingContinuationWith passContinuation: @Sendable @escaping (CheckedContinuation<Void, Never>) async -> Void
+    ) {
```
MahdiBM, line 22: Some explanation.

Tests/PennyTests/Fake/FakeMainService.swift
```diff
@@ -23,7 +25,7 @@ actor FakeMainService: MainService {
         cacheStorage.guilds[TestData.vaporGuild.id] = TestData.vaporGuild
         self.cache = await DiscordCache(
             gatewayManager: manager,
-            intents: [.guilds, .guildMembers],
+            intents: [.guilds, .guildMembers, .messageContent, .guildMessages],
```
MahdiBM, line 28: Make it the same as the real app.

Tests/PennyTests/Tests/GatewayProcessingTests.swift
```diff
@@ -1,33 +1,97 @@
...
-            Task {
+            mainServiceTask = Task<Void, any Error> {
                 try await Penny.start(mainService: fakeMainService)
             }
             await fakeMainService.waitForStateManagerShutdownAndDidShutdownSignals()
         }
+
+        deinit {
+            mainServiceTask.cancel()
+        }
```
MahdiBM, line 40: We cancel the task so it's not really unstructured.

Sources/Penny/MainService/PennyService.swift
```diff
@@ -147,30 +164,59 @@ struct PennyService: MainService {
...
+                await context.botStateManager.addCachesPopulationContinuation($0)
+            }
+        )
+        let swiftReleasesCheckerWrappedService = WaiterService(
+            underlyingService: context.swiftReleasesChecker,
+            processingOn: context.backgroundProcessor,
+            passingContinuationWith: {
+                await context.botStateManager.addCachesPopulationContinuation($0)
+            }
+        )
```
MahdiBM, line 206: See comments as to why we're using "wrapped" services.
The `WaiterService` type also has some doc comments.

Tests/PennyTests/Tests/GatewayProcessingTests.swift
```diff
@@ -1,33 +1,97 @@
...
+        let runningService = Task<Void, Never> {
+            try! await wrappedService.run()
+        }
+
+        try await Task.sleep(for: .seconds(5))
+
+        runningService.cancel()
+
+        #expect(await sampleService.didRun == false)
+    }
```
MahdiBM, line 93: Some tests for the waiter service

Tests/PennyTests/Tests/GatewayProcessingTests.swift
```diff
@@ -358,35 +425,43 @@ extension SerializationNamespace.GatewayProcessingTests {
...
+
+        serviceTask.cancel()
     }
 
     @Test
     func swiftReleasesChecker() async throws {
-        context.services.swiftReleasesChecker.run()
+        let serviceTask = Task<Void, any Error> { [self] in
+            try await self.context.swiftReleasesChecker.run()
+        }
```
MahdiBM, line 464: This `Task` isn't really "unstructured" since we cancel it in the end.

Tests/PennyTests/Tests/GatewayProcessingTests.swift
0xTim: This would automatically get cancelled on deinit right?

Sources/Penny/BotStateManager.swift
```diff
@@ -25,97 +26,129 @@ import Logging
...
+    private func shutdown() async {
+        await context.cachesService.gatherCachedInfoAndSaveToRepository()
+        await send(.didShutdown)
+        self.canRespond = false
 
-            try await Task.sleep(for: disableDuration)
-            await startAllowingResponses()
-            logger.critical("AWS has not yet shutdown this instance of Penny! Why?!")
+        guard (try? await Task.sleep(for: disableDuration)) != nil else {
+            return /// cancelled
```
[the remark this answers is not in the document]
MahdiBM: Changed the code a little bit, i think this is more close to what i meant to do:
https://github.com/vapor/penny-bot/commit/6d3485ae1b4c9124c1955db8d6f5edf8bc1c7835

For example if the background processor is cancelled, then the task can get cancelled and the Task.sleep will throw?

I should go double check my knowledge but that's what I'm thinking right now.

Sources/Penny/HandlerContext.swift
```diff
@@ -1,22 +1,50 @@
 import Rendering
 import Shared
+import ServiceLifecycle
 
-struct HandlerContext: Sendable {
+final class HandlerContext: Sendable {
+    /// Contain references to this class so need to be initialized separately
+    nonisolated(unsafe) var botStateManager: BotStateManager!
```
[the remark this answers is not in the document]
MahdiBM: Because it is only ever modified once, and we know it won't cause anything bad.

MahdiBM:
@0xTim 
> This would automatically get cancelled on deinit right?

Is that a fact? i guess it would make sense. I couldn't find anything in the proposals and haven't manually tried it yet.

Sources/Penny/BotStateManager.swift
```diff
@@ -25,97 +26,129 @@ import Logging
...
+            self.context.backgroundProcessor.process {
+                await self.cancelIfCachePopulationTakesTooLong()
+            }
+            self.context.backgroundProcessor.process {
+                await self.send(.shutdown)
+            }
+        }
+
+        /// Wait indefinitely
+        let (stream, _) = AsyncStream.makeStream(of: Void.self)
```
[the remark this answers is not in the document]
MahdiBM: Actually took the code from Franz 😅

We want `BotStateManager` to be a `Service`, but the type itself doesn't do any long-running work.
So we have to somehow make it seem like we're doing some long running work and make sure the `run()` method doesn't exit.
If the `run()` method exits, `ServiceLifecycle` will think something has gone wrong and will shutdown the whole app, gracefully.
There is a setting to configure this behavior in `ServiceLifecycle`, but it's a per-`ServiceLifecycle` setting not a per-Service setting, so if we change that then even if another service's `run()` method finishes abruptly, we won't the desired behavior of `ServiceLifecycle` gracefully shutting down everything.

Tests/PennyTests/Tests/GatewayProcessingTests.swift
MahdiBM: Is that a fact? I guess it would make sense. I couldn't find anything in the proposals and haven't manually tried it yet.

I removed that deinit now.

0xTim: I'm pretty sure otherwise it just wouldn't deinit right?

Sources/Penny/BotStateManager.swift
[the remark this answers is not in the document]
0xTim: Seems like a gap in ServiceLifecycle but ok!

Tests/PennyTests/Tests/GatewayProcessingTests.swift
MahdiBM: nope.
```swift
@main
struct EntryPoint {
    static func main() async throws {
        print("started")
        let task = Task {
            if (try? await Task.sleep(for: .seconds(60))) == nil {
                print("cancelled")
                fflush(stdout)
            }
        }
        print("over: \(task)")
    }
}
```
results in:
```
started
over: Task<(), Never>(_task: (Opaque Value))
Program ended with exit code: 0
```

MahdiBM: to be clear, adding `task.cancel(); try await task.value` does make it print "cancelled".

Sources/Penny/BotStateManager.swift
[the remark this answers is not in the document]
MahdiBM: Changed it to `return /// Somewhere else cancelled the Task`

0xTim (approved):

PR #258, opened 2024-10-25
Better Evolution Proposal error handling

MahdiBM:

### DIFF
--- a/Sources/Penny/EvolutionChecker.swift
+++ b/Sources/Penny/EvolutionChecker.swift
@@ -18,6 +18,8 @@ actor EvolutionChecker {
 
     var storage = Storage()
 
+    var reportedProposalIDsThatContainErrors: Set<String> = []
+
     /// The minimum time to wait before sending a queued-proposal
     let queuedProposalsWaitTime: Double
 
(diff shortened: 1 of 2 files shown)


Sources/Penny/EvolutionChecker.swift
```diff
@@ -57,6 +59,24 @@ actor EvolutionChecker {
             return
         }
 
+        let proposalIDsWithError = proposals.filter {
+            !($0.errors ?? []).isEmpty
+        }.map(\.id)
+        if !proposalIDsWithError.isEmpty {
+            let allAreAlreadyReported = proposalIDsWithError.allSatisfy {
+                self.reportedProposalIDsThatContainErrors.contains($0)
+            }
```
gwynne, line 68:
```suggestion
            let allAreAlreadyReported = proposalIDsWithError.isSuperset(of: self.reportedProposalIDsThatContainErrors)
```

gwynne (changes requested):
One efficiency nit.

Sources/Penny/EvolutionChecker.swift
MahdiBM:
```swift
self.reportedProposalIDsThatContainErrors
    .isSuperset(of: proposalIDsWithError)
```
is what we'd want here 🙂

gwynne: What fun is it if I do _all_ the work for you? It can't at all be that I made a mistake! Or that I meant `isSubsetOf`! Can't be! 🤣

gwynne (approved):

PR #277, opened 2024-12-20
Update GitHub OpenAPI with idiomatic namings

MahdiBM:


Lambdas/GitHubAPI/openapi-generator-config.yml
```diff
@@ -2,3 +2,4 @@ generate:
   - types
   - client
 accessModifier: package
+namingStrategy: idiomatic
```
MahdiBM, line 5: The new setting

gwynne (approved):
Not sure what problem this actually solves, but sure, go ahead 😆

Lambdas/GHHooks/EventHandler/Handlers/DocsIssuer.swift
```diff
@@ -65,28 +65,28 @@ struct DocsIssuer {
     }
 
     func getPRsRelatedToCommit() async throws -> [SimplePullRequest] {
-        try await context.githubClient.repos_list_pull_requests_associated_with_commit(
+        try await context.githubClient.reposListPullRequestsAssociatedWithCommit(
             path: .init(
                 owner: self.repo.owner.login,
                 repo: self.repo.name,
-                commit_sha: self.commitSHA
+                commitSha: self.commitSHA
```
0xTim, line 72: Now we just need to capitalise this properly 😛

0xTim (approved):

Lambdas/GHHooks/EventHandler/Handlers/DocsIssuer.swift
MahdiBM: I don't think it's possible, automatically, but the generator does take custom key value name mapping pairs and we could use that although I don't care enough for that.

PR #353, opened 2025-05-30
Enable `InternalImportsByDefault` feature flag

MahdiBM:


.sourcekit-lsp/config.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://raw.githubusercontent.com/swiftlang/sourcekit-lsp/refs/heads/main/config.schema.json",
```
MahdiBM, line 2: mostly unrelated change: Add the schema file so VSCode can show if you're using an invalid JSON field.

.swift-version
```diff
@@ -0,0 +1 @@
+6.1.2
```
MahdiBM, line 1: mostly unrelated change: Swiftly to use 6.1.2

.vscode/settings.json
```diff
@@ -81,5 +81,6 @@
   "terminal.integrated.scrollback": 10000,
   "workbench.editor.enablePreview": false,
   "yaml.format.enable": true,
-  "yaml.maxItemsComputed": 1000000
+  "yaml.maxItemsComputed": 1000000,
+  "swift.debugger.debugAdapter": "CodeLLDB"
```
MahdiBM, line 85: The new LLDB-DAP that is in the toolchains is a bit buggy, so set this to use the good ol' CodeLLDB for now.

PR #357, opened 2025-06-05
Better release branch detection

MahdiBM:
`repoName-num` looks to be the crowd favorite for naming release branches so let's accept it.
Examples: `vapor-5`, `console-kit-5`.


0xTim:
What's the usage of this way btw?

PR #367, opened 2025-08-27
Simpler Docker installation in CI

MahdiBM:


.github/workflows/deploy-penny.yml
```diff
@@ -51,12 +51,7 @@ jobs:
           key: "penny-release-static-build-${{ runner.os }}-${{ github.event.after }}"
 
       - name: Install Docker
-        run: |
-          # Make scripts executable
-          chmod +x ./.github/scripts/*
-
-          # Install Docker
-          ./.github/scripts/install-docker.bash
+        run: curl -fsSL https://get.docker.com | bash
```
MahdiBM, line 54: Until a few weeks ago I wasn't aware such a script exists.

PR #382, opened 2026-04-27
Update dependencies + improve CI flags

MahdiBM:
~~https://github.com/swiftlang/swift-syntax/issues/3317 is fixed in the latest swift-syntax, so we can remove the "disable prebuilts" flag from Penny's CI.~~ Still getting meaningless errors. Looks like a prebuilts issue, not swift-syntax.
I've also manually added the default extra-flags just so it's clear what flags are being used.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
       with_tsan: false
       with_api_check: false
       with_linting: true
-      extra_flags: "--disable-experimental-prebuilts"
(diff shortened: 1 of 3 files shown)


gwynne:
IIRC you actually *have* to add the default flags manually if you specify `extra_flags` at all, the idea being that that lets you remove the default ones if you need to.

gwynne (approved):

MahdiBM:
@gwynne right. That comment was from before I had to re-add "disable prebuilts".

PR #393, opened 2026-06-16
Fix the GH Sponsors Lambda

0xTim:
There were some missing environment variables meaning the lambda was never working. Add them back


MahdiBM (approved):
This is good as-is, but IIRC other functions also have some env vars. Not sure, can't check for a bit.

MahdiBM:
> meaning the lambda was never working

Hmm i'd think thats not true in Vapor's setup, but again, I'l have to check later.
Possibly at some point something got messed up or something.

0xTim:
> Hmm i'd think thats not true in Vapor's setup, but again, I'l have to check later.
> Possibly at some point something got messed up or something.

The sponsors lambda hasn't worked for years, but there's more to do. Looking into it

0xTim:
@MahdiBM ok so when we migrated to the new Penny architecture, the sponsorship event was left to implement with just the log. So I ended up implementing it alongside everything else, feel free to take a look!

gwynne (approved):

PR #417, opened 2026-08-12
Remove the dependency on Foundation

MahdiBM:
Some of our dependencies rely on Foundation so we can't go Foundation-free this easily, but let's at least be ready and ensure Penny itself contains no Foundation usages.


Sources/Shared/+String.swift
```diff
@@ -1,11 +1,64 @@
-/// Import full foundation even on linux for `addingPercentEncoding`, for now.
-import Foundation
-
 extension String {
     package func urlPathEncoded() -> String {
-        self.addingPercentEncoding(
-            withAllowedCharacters: .urlPathAllowed
-        ) ?? self
+        self.percentEncoded(
```
MahdiBM, line 54: Pretty much all actual changes are in this file.
These are not super performant although they do fine, but I didn't care, they should be plenty.

.github/workflows/deploy-all-lambdas.yml
```diff
@@ -19,10 +19,12 @@ jobs:
       AWS_DEFAULT_REGION: eu-west-1
 
     steps:
-      - name: Install AWSCLI v2, jq, zstd, zip, binutils
+      - name: Install AWSCLI v2, jq, zstd, and stub docker for the lambda-build plugin
         run: |
           dnf update -y
-          dnf install jq awscli zstd zip unzip binutils -y
+          dnf install jq awscli zstd -y
```
MahdiBM: Apparently `lambda-build` requires docker even if it's not using docker anyway and building on this container itself. Should file an issue for it.

.github/workflows/deploy-all-lambdas.yml
```diff
@@ -56,29 +58,11 @@ jobs:
           for name in ${{ steps.find_package_names.outputs.names }}; do
             swift package \
               --allow-network-connections docker \
-              archive \
+              lambda-build \
               --verbose \
-              --base-docker-image swift:6.3-amazonlinux2023 \
               --output-path ./zips \
+              --no-strip \
```
MahdiBM, line 64: This is intentional. I want to see where we can get without stripping the debug/backtrace-related info.

Lambdas/GHHooks/Extensions/+PR.swift
```diff
@@ -73,14 +72,7 @@ extension String {
     fileprivate var hasDoNotMergePrefix: Bool {
         let folded = self.lowercased()
             .filter { !$0.isPunctuation }
-            .folding(
-                options: .caseInsensitive,
-                locale: nil
-            )
```
MahdiBM, line 79: this is redundant, there is a ".lowercased()" right above.

0xTim (approved):

PR #420, opened 2026-08-19
Implement Spam-guard

MahdiBM:
Implements guarding the Discord server from being spammed by the same user (multiple messages in a few seconds in multiple channels).

This behavior is very uncommon by an actual human. I don't recall seeing it even once, so yeah...

Deleted all messages of the spammer in the last some-seconds (25 right now) and times them out for 6 hours until a moderator can make a decision. Logs the deleted messages so if they are deleted in mistake, we can at least manually funnel the messages to the author so they can at least repost it.

Almost all users who have any roles are exempt from all this.

Currently doesn't care if messages contain the same content, but we could also check for that in the future.

### DIFF
--- a/Sources/Penny/Handlers/SpamHandler/SpamCache.swift
+++ b/Sources/Penny/Handlers/SpamHandler/SpamCache.swift
@@ -1,0 +1,116 @@
+import Algorithms
+import DiscordBM
+import Shared
+
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
+import Foundation
+#endif
+
+/// Cache for detecting users who spam multiple channels at once.
+actor SpamCache<ClockType: Clock> where ClockType.Instant.Duration == Swift.Duration {
+
+    /// Only what is needed to detect the spam and to find the messages back in the Discord cache.
+    struct Entry: Sendable {
+        let messageId: MessageSnowflake
+        let channelId: ChannelSnowflake
+        let timestamp: ClockType.Instant
+
+        init(message: Gateway.MessageCreate, now: ClockType.Instant) {
+            self.messageId = message.id
+            self.channelId = message.channel_id
+            let lag = Duration.seconds(Date.now.timeIntervalSince(message.timestamp.date))
+            self.timestamp = now.advanced(by: .zero - lag)
+        }
+    }
+
+    enum Verdict: Sendable {
+        case notSpam
+        case spam(entries: [Entry])
+        case alreadyFlagged
+    }
+
+    /// How far back a message is taken into account.
+    static var window: Duration { .seconds(20) }
+    /// How many distinct channels a user must post in, in a `window`, to be considered a spammer.
+    static var channelThreshold: Int { 3 }
+    /// Messages that arrive in this period after a flag are removed without reporting again.
+    static var flagDuration: Duration { .seconds(60) }
+
+    private let clock: ClockType
+    private var entries: [UserSnowflake: [Entry]] = [:]
+    private var flagged: [UserSnowflake: ClockType.Instant] = [:]
+
+    init(clock: ClockType, runCleanupBackgroundTask: @escaping (@escaping @Sendable () async -> Void) -> Void) {
+        self.clock = clock
+        runCleanupBackgroundTask {
+            await self.cleanupPeriodically()
+        }
+    }
+
+    func recordAndCheck(
+        _ message: Gateway.MessageCreate,
+        of userId: UserSnowflake
+    ) -> Verdict {
+        let now = self.clock.now
+
+        if let flaggedAt = self.flagged[userId],
+            flaggedAt.duration(to: now) <= Self.flagDuration
+        {
+            return .alreadyFlagged
+        }
+
+        let newEntry = Entry(message: message, now: now)
+
+        var recentMessages = self.entries.removeValue(forKey: userId) ?? []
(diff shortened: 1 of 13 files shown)


Sources/Penny/Handlers/SpamHandler/SpamCache.swift
```diff
@@ -0,0 +1,116 @@
...
+        case spam(entries: [Entry])
+        case alreadyFlagged
+    }
+
+    /// How far back a message is taken into account.
+    static var window: Duration { .seconds(20) }
+    /// How many distinct channels a user must post in, in a `window`, to be considered a spammer.
+    static var channelThreshold: Int { 3 }
+    /// Messages that arrive in this period after a flag are removed without reporting again.
+    static var flagDuration: Duration { .seconds(60) }
```
MahdiBM, line 39: 3 messages in 20 seconds, and for 60 after if that happens, by 1 user.

See the `SpamCache` tests if you want to check the behavior more.

Sources/Penny/Handlers/SpamHandler/SpamCache.swift
```diff
@@ -0,0 +1,116 @@
...
+    private let clock: ClockType
+    private var entries: [UserSnowflake: [Entry]] = [:]
+    private var flagged: [UserSnowflake: ClockType.Instant] = [:]
+
+    init(clock: ClockType, runCleanupBackgroundTask: @escaping (@escaping @Sendable () async -> Void) -> Void) {
+        self.clock = clock
+        runCleanupBackgroundTask {
+            await self.cleanupPeriodically()
+        }
+    }
```
MahdiBM, line 50: Just taking a closure made it easier to mock it in tests although optimally we'd have `BackgroundProcessorProtocol` instead of just a `BackgroundProcessor`.

Sources/Penny/Handlers/SpamHandler/SpamCache.swift
```diff
@@ -0,0 +1,116 @@
...
+
+    func recordAndCheck(
```
MahdiBM, line 52: Most logic is here, again I'd say you should look at the tests instead.

Sources/Penny/Handlers/SpamHandler/SpamHandler.swift
```diff
@@ -0,0 +1,169 @@
...
+            author.bot != true,
+            author.id != Constants.botId,
+            event.guild_id == Constants.vaporGuildId
+        else { return }
+
+        for role in event.member?.roles ?? [] {
+            if Constants.Roles.elevatedPublicCommandsAccessSet.contains(role) {
+                return
+            }
+        }
```
MahdiBM, line 43: Most people who have any rules are exempt.

Sources/Penny/Handlers/SpamHandler/SpamHandler.swift
```diff
@@ -0,0 +1,169 @@
...
+                messageId: event.id,
+                channelId: event.channel_id,
+                reason: Self.reason
+            )
+        case .spam(let entries):
+            await discordService.timeoutMember(
+                userId: author.id,
+                for: Self.timeoutDuration,
+                reason: Self.reason
+            )
```
MahdiBM, line 59: Timeout for 6 hours until a moderator can check.

Sources/Penny/Handlers/SpamHandler/SpamHandler.swift
```diff
@@ -0,0 +1,169 @@
...
+            await discordService.sendMessage(
+                channelId: Constants.Channels.modLogs.id,
+                payload: self.makeMessagePayload(
+                    spamEntries: entries,
+                    messages: cachedMessages,
+                    lastMessage: event,
+                    author: author
+                )
+            )
+            await self.removeMessages(of: entries, and: cachedMessages)
```
MahdiBM, line 71: Remove all their messages in the "window" + 5 seconds (currently 25).

Sources/Penny/Services/DiscordService/DiscordService.swift
```diff
@@ -448,9 +521,9 @@ actor DiscordService {
     }
 
     func memberHasRolesForElevatedPublicCommandsAccess(member: Guild.Member) -> Bool {
-        Constants.Roles.elevatedPublicCommandsAccess.contains(where: {
-            member.roles.contains($0.rawValue)
-        })
+        !Set(member.roles)
+            .intersection(Constants.Roles.elevatedPublicCommandsAccessSet)
+            .isEmpty
     }
```
MahdiBM, line 527: same functionality

Tests/PennyTests/Tests/SpamCacheTests.swift
```diff
@@ -0,0 +1,243 @@
...
+        for index in 0..<2 {
+            let message = makeMessage(id: try .makeFake(), inChannel: index, of: spammer)
+            let verdict = await cache.recordAndCheck(message, of: spammer)
+            #expect(verdict.isNotSpam, "\(verdict)")
+            clock.advance(by: .seconds(5))
+        }
+    }
+
+    @Test
+    func threeChannelsInTheWindowAreSpam() async throws {
```
MahdiBM, line 52: Basic test that it works a bit at least. More tests above and below.

0xTim (approved):
LGTM

